PR #10, opened 2022-02-12
Adding Super Mario 64 and VVVVVV

ScootyPuffJr1:


Berserker66:
I think the idea was to make the SM64 stars local, if I recall.

ScootyPuffJr1:
I argued that they should be local to avoid comparisons to LttP's non-local triforce hunts, but others claimed they aren't comparable. Forcing stars local leaves about 5 checks and really doesn't make sense to have in a multiworld, in my opinion. I'm conflicted because non-local stars goes against suggested goal types for the async, but local stars seem pointless. I'm willing to try non-local once and see how it goes.

ScootyPuffJr1:
Updated game weights, changed random setting for Timespinner to produce parseable results, updated SM64 an V6 to 0.2.5, made Power Stars local.

PR #13, opened 2022-06-04
dont force power stars local for SM64

t3hf1gm3nt:
no one wants to play single player sm64 for a big multiworld

### DIFF
--- a/async_weights.yaml
+++ b/async_weights.yaml
@@ -951,8 +951,6 @@ Super Mario 64:
   BuddyChecks:
     false: 1
     true: 1
-  local_items:
-    ["Power Star"]
 VVVVVV:
   DoorCost: 
     3: 1 


Berserker66:
should be discussed first, as the change was made as a result of a discussion.

t3hf1gm3nt:
we had the discussion about how sm64 filler seeds should be months ago, and im 99% certain that it was determined that until the star door requirements can be set lower than 50, that forcing local power stars would be unreasonable. im trolling through the discord now to find the relevant conversation

t3hf1gm3nt:
actually here: https://github.com/ArchipelagoMW/AsyncTools/pull/10 you already had discussed this once with scooty when he made them local in the first place, and scooty was also conflicted. i guess we do need to have a rehash of this discussion with the community because a lot of things changed with sm64 since then

Joethepic:
based on the discussion most people be on the side of non local

t3hf1gm3nt:
thats what im getting from the conversation as well

t3hf1gm3nt:
closing this PR because The T has opened one using the new restructure

PR #16, opened 2022-06-04
change minecraft settings

Joethepic:
made advancements not the same and bosses not the same

### DIFF
--- a/async_weights.yaml
+++ b/async_weights.yaml
@@ -384,11 +384,13 @@ Minecraft:
   include_insane_advancements: off
   include_postgame_advancements: off
   shuffle_structures: on
+  structrue
   required_bosses:
     none: 50
     ender_dragon: 50 
     wither: 50
     both: 10
+  structure_compasses: on
 A Link to the Past:
   ### Logic Section ###
   # Warning: overworld_glitches is not available and minor_glitches is only partially implemented on the door-rando version
--- a/async_weights.yaml
+++ b/async_weights.yaml
@@ -384,7 +384,6 @@ Minecraft:
   include_insane_advancements: off
   include_postgame_advancements: off
   shuffle_structures: on
-  structrue
   required_bosses:
     none: 50
     ender_dragon: 50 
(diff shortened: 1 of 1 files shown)


Kippi00:
I would argue perhaps using random-range for advancements, as with random there is a non-zero chance that you could need 0 advancements to win and a non-zero chance you could need 95 advancements to win. 0 advancements to win doesn't seem async friendly, IMO 

There are 107 advancements in the current version of Minecraft in AP and considering the filler yaml excludes postgame and unreasonable advancements, that's at most 14 advancements that's excluded, leaving 93 that can be expected to be done, leaving room for 2 of those postgame/unreasonable advancements to be needing to be done

Perhaps something like a random ranging from 30 to 80 or 40 to 90 would be better

Joethepic:
> I would argue perhaps using random-range for advancements, as with random there is a non-zero chance that you could need 0 advancements to win and a non-zero chance you could need 95 advancements to win. 0 advancements to win doesn't seem async friendly, IMO
> 
> There are 107 advancements in the current version of Minecraft in AP and considering the filler yaml excludes postgame and unreasonable advancements, that's at most 14 advancements that's excluded, leaving 93 that can be expected to be done, leaving room for 2 of those postgame/unreasonable advancements to be needing to be done

> Perhaps something like a random ranging from 30 to 80 or 40 to 90 would be better

according to the template

Joethepic:
> I would argue perhaps using random-range for advancements, as with random there is a non-zero chance that you could need 0 advancements to win and a non-zero chance you could need 95 advancements to win. 0 advancements to win doesn't seem async friendly, IMO
> 
> There are 107 advancements in the current version of Minecraft in AP and considering the filler yaml excludes postgame and unreasonable advancements, that's at most 14 advancements that's excluded, leaving 93 that can be expected to be done, leaving room for 2 of those postgame/unreasonable advancements to be needing to be done
> 
> Perhaps something like a random ranging from 30 to 80 or 40 to 90 would be better

this should be fixed i made it 30-80 assuming i did it right

Kippi00:
Yep that looks right to me.

PR #17, opened 2022-06-04
Added SC2 to async filler yaml

TheCondor07:
Randoms between casual and normal difficulty.  All other options default

### DIFF
--- a/async_weights.yaml
+++ b/async_weights.yaml
@@ -16,6 +16,7 @@ game:
   VVVVVV: 20
   Meritous: 10
   SMZ3: 30
+  Starcraft 2 Wings of Liberty: 20
 Timespinner:
   StartWithJewelryBox: true
   DownloadableItems: true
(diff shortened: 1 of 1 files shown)


Berserker66:
no random mission order at all?

TheT113:
I initially thought the same thing, Berserker, and I did wonder if; anyone who feels confident enough to take random mission order is probably going to use one of their custom slots on it? but I also think a lower divide, like 4 off 1 on would probably be fine?

TheCondor07:
Yeah, changed it for a low chance of random mission order.  My concern was for newer players to the series and I will admit I have not gotten enough feedback on random mission order for how much harder it makes the game.

t3hf1gm3nt:
i dont think it makes it much easier/more difficult with mission rando unless you get a very unlucky set of missions, but rando gonna rando. definitely feel that there should at least be some fillers that are mission rando'd, but the majority should be vanilla since asyncs are prime targets for newer players to games. i think the 40/10 split is fine

PR #19, opened 2022-06-05
Modularize Mystery Settings for Async Filler Seeds

ThePhar:
Something about having all the game settings in 1 master file just irks me the wrong way, so I've broken it up, created a generator for creating a completed `.yaml` file, and added a read me on how it works.

~~Still need to make tweaks for games that don't generate on current release, but should be a good proof of concept while I'm working on it.~~ Apparently it just works.

Screenshot of the generator in action:
image

Looks good to the website mystery check too:
image


ThePhar:
I kept thinking about doing this, and now I regret not doing it until today since there's a bunch of PRs open. Oof.

ThePhar:
Some simple error checking as well, I can showcase.

Game settings file does not exist, but game is specified in __meta__:
image

Game settings file exists, but top-level key does not match the Game name.
image
image

Game settings file exists, but there are additional top-level keys specified in addition to <game> key:
image
image

Berserker66 (approved):

PR #24, opened 2022-06-09
bettering accessability to mario 

Joethepic:


TheT113:
why would we want to turn off strict cap requirements? that would make things harder and less accessible.

Joethepic:
I meant to make it the opposite I’ll fix that

Ijwu (changes requested):
Your PR deletes the minecraft yaml. It's not getting merged until that change gets reverted, probably.

Joethepic:
fixed

ScootyPuffJr1:
Why do we want to completely remove the possibility of false cap requirements? Some people might want that option on.

Joethepic:
I believe we should turn it off @ScootyPuffJr1  because it makes things more accessible and more likely for seeds to get picked up for this kind of logic change I believe it is right to always keep it the easiest much like keeping lamp logic on keeping glitchless in oot and lttp.

ScootyPuffJr1:
These filler seeds are not meant to always be easy. We frequently get new async players that are not new to the games they are playing. Other game settings are definitely not always set to be easy. Examples: LttP item pool hard and functionality hard and expert, Rogue Legacy has the chance of challenge bosses, SM allows veteran preset and up to hardcore difficulty. In my opinion, we should lean toward easy but still allow more difficult seeds to happen.

Berserker66 (approved):

PR #27, opened 2022-06-19
Update __meta__.yaml: Game weights

ScootyPuffJr1:
Updating game weights

### DIFF
--- a/games/__meta__.yaml
+++ b/games/__meta__.yaml
@@ -7,16 +7,16 @@ requires:
 # Define your game here!
 game:
   A Link to the Past: 150
-  Factorio: 20
-  Minecraft: 50
-  Slay the Spire: 20
-  Risk of Rain 2: 30
+  Factorio: 15
+  Minecraft: 60
+  Slay the Spire: 15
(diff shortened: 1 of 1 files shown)


black-sliver:
I believe we wanted to remove subnautica if it does not find a maintainer? 
Not sure about RL vs. SM. Just looked at /stats and SM seems more popular.

Berserker66:
Subnautica can stay for now, but at some point we need to cut out the server-side compatibility code.

ScootyPuffJr1:
SM definitely has more people than RL (269-156). I wasn't sure if the addition of SMZ3 reduced the number of people willing to play just SM. Both SM and RL had their filler seeds claimed very quickly in the last async.

black-sliver:
```
  Ocarina of Time: 50
  # ...
  Super Metroid: 40
  Rogue Legacy: 60
```
Just feels odd, but if you think your numbers are worth testing, I'm fine with it.

Last async had very few fillers in general. I hope we'll get a higher success rate and more slots this time :)

PR #31, opened 2022-09-22
SA2: Add new options

TheT113:
I don't feel strongly about the junk fill and trap fill, or whether traps should be included at all; I could see changing random-low to a random-range with lower values as well, but also don't feel particularly strongly about it. Anyone else is free to make that suggestion or change though.

I wanted a 50/50 chance of Chao Garden, with a leaning towards more of it if it gets put in; Chao Karate less so, because it requires the DLC. So including whether Chao Karate is on or off should be expected on the spreadsheet.

Setting rank to E to encourage new players to take these worlds; an experienced player will probably be making a custom. Could still see this change though.

### DIFF
--- a/games/Sonic Adventure 2 Battle.yaml
+++ b/games/Sonic Adventure 2 Battle.yaml
@@ -22,7 +22,7 @@ Sonic Adventure 2 Battle:
     beginner: 1
     intermediate: 1
     expert: 3
-  include_chao_karate: # 
+  include_chao_karate:
     false: 2
     true: 1 
   chao_race_checks: all
--- a/games/Sonic Adventure 2 Battle.yaml
+++ b/games/Sonic Adventure 2 Battle.yaml
@@ -13,3 +13,22 @@ Sonic Adventure 2 Battle:
     "4": 2
     "5": 3
   level_gate_distribution: random
+  required_rank: e
+  level_gate_cost:
+    medium: 1
+    high: 2
+  chao_garden_difficulty:
+    none: 5
+    beginner: 1
+    intermediate: 1
+    expert: 3
+  include_chao_karate: # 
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
Would like @PoryGone and anyone else who actually knows SA2 to review this please

games/Sonic Adventure 2 Battle.yaml
```diff
@@ -13,3 +13,22 @@ Sonic Adventure 2 Battle:
...
   level_gate_distribution: random
+  required_rank: e
+  level_gate_cost:
+    medium: 1
+    high: 2
+  chao_garden_difficulty:
+    none: 5
+    beginner: 1
+    intermediate: 1
+    expert: 3
```
t3hf1gm3nt, line 24: so i do know next to nothing about SA2, but just going by those names, i would think weights should be something more like 5:3:1:1? why is there a bigger chance for expert more than the less difficult options is what im asking really

smahlt:
> Setting rank to E to encourage new players to take these worlds; an experienced player will probably be making a custom. Could still see this change though.

I feel this statement contradicts the 10% chance of expert chao races and 33% of chao karate being enabled, unless you expect "new players" (your definition of this term also matters) to use the animal glitch. At the same time, I don't know how many SA2 filler seeds would get rolled per async, so maybe 10% is high enough to make it actually appear every now and then.

TheT113:
I'm not sure what your question is.

Taking the time to replay levels to collect the easily accessible items needed to raise chao stats (or using the rings to buy chao fruit, which is much faster) doesn't have anything to do with difficulty.

PoryGone:
I fully agree with rank requirement just being `E` for all filler slots.

PR #32, opened 2022-09-22
Hollow Knight yaml

SnowOfAllTrades:
Hollow Knight async filler yaml

### DIFF
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -101,7 +101,3 @@ Hollow Knight:
     - King_Fragment
     - Queen_Fragment
     - Void_Heart
-
-description: 'Hollow Knight Async Filler yaml'
-game: Hollow Knight
-name: Player{number}
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -1,5 +1,4 @@
 Hollow Knight:
-  progression_balancing: 0
   accessibility: items
   RandomizeDreamers: 'true'
   RandomizeSkills: 'true'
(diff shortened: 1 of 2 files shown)


TheT113:
You'll want to remove the last 3 lines, or else it won't function correctly.

SnowOfAllTrades:
done, thanks

games/Hollow Knight.yaml
```diff
@@ -0,0 +1,103 @@
+Hollow Knight:
+  progression_balancing: 0
```
t3hf1gm3nt, line 2: prog balancing is unnecessary since it gets meta'd to off anyway

SnowOfAllTrades: Fair enough, I'll remove the line

ScootyPuffJr1:
You need to add Hollow Knight to the __meta__.yaml file, too.

games/Hollow Knight.yaml
```diff
@@ -0,0 +1,102 @@
...
+  SplitMothwingCloak: 'random'
+  SplitMantisClaw: 'random'
+  CostSanity: off
+  CostSanityHybridChance: 10
+  CostSanityEggWeight: 1
+  CostSanityGrubWeight: 1
+  CostSanityEssenceWeight: 1
+  CostSanityCharmWeight: 1
+  CostSanityGeoWeight: 8
+  local_items:
```
alwaysintreble, line 96: uhhh have you test generated with this? the way you have this formatted they're all separate lists which should break. remove the `-` on all but the first line. they still need to be indented though.

PR #33, opened 2022-09-22
Update Subnautica with new settings

SnowOfAllTrades:
Adding new settings for Creature Scans (amount) and Creature Scan logic, as well as a little bit of variety to the Item Pool setting. Goal settings stay the same.

### DIFF
--- a/games/Subnautica.yaml
+++ b/games/Subnautica.yaml
@@ -1,9 +1,20 @@
 Subnautica:
+  progression_balancing: 0
+  accessibility: items
   item_pool:
-    valuable
+    standard: 1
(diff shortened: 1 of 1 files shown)


games/Subnautica.yaml
```diff
@@ -1,9 +1,17 @@
 Subnautica:
+  progression_balancing: 0
```
t3hf1gm3nt, line 2: prog balancing is unnecessary line since it gets meta'd to off anyway

SnowOfAllTrades: removed it

PR #34, opened 2022-09-27
[LTTP] Remove some more difficult settings

t3hf1gm3nt:
Removes weights from insanity ER, moving them to full, and expert item pool/functionality and moves them to hard. Was discussed that these settings are a bit much for a filler yaml and should be left for people to select for customs if they wish.

### DIFF
--- a/games/A Link to the Past.yaml
+++ b/games/A Link to the Past.yaml
@@ -176,7 +176,7 @@ A Link to the Past:
   glitch_boots:
     on: 50 # Start with Pegasus Boots in any glitched logic mode that makes use of them
     off: 0
-triggers:
+  triggers:
   #Turn killable thieves on if enemy shuffle is on
   - option_name: enemy_shuffle
     option_category: A Link to the Past
--- a/games/A Link to the Past.yaml
+++ b/games/A Link to the Past.yaml
@@ -53,9 +53,9 @@ A Link to the Past:
     dungeonscrossed: 10
     simple: 1 # Entrances are grouped together before being randomized. Simple uses the most strict grouping rules
     restricted: 1 # Less strict than simple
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
Changes seem fine to me.

alwaysintreble:
killable thieves should probably be a trigger and ganon_ped should be removed since we have local_ganon_ped. i'd also like to see some shop shuffle options but that might just be me

t3hf1gm3nt:
I'll admit all these things are outside the scope of my original intention for this update but I'll answer my thoughts anyway.

Killable thieves as a trigger to only be on when enemizer is on? That's something I could agree too.

What do you mean by local Ganon ped? Ganon ped is always local since dungeon prizes are always local

I never play with shop shuffle so I'd have no idea what good options would be for those. I'll leave that to someone with better experience with it to submit a PR, but I could be convinced to have some options in there for it

alwaysintreble:
i got confused with tfh 🤦

TheT113:
I like all of these changes. 👍

games/A Link to the Past.yaml
```diff
@@ -176,3 +176,11 @@ A Link to the Past:
   glitch_boots:
     on: 50 # Start with Pegasus Boots in any glitched logic mode that makes use of them
     off: 0
+triggers:
```
ThePhar, line 179: This needs to be at the same level as all the other options or it will fail to generate a yaml.

ThePhar (changes requested):

games/A Link to the Past.yaml
t3hf1gm3nt: indented over. look good now?

ScootyPuffJr1:
Triggers look good

PR #35, opened 2022-10-01
add scripts to make mass generation easier

alwaysintreble:
these files currently assume they're placed in the root Archipelago directory but make mass world generation/testing easier


ThePhar:
Mind adding to the readme a blurb what each batch file does?

PR #36, opened 2022-10-01
Update The Witness.yaml

Beaker271:
Updated the witness yaml for the current version and included new settings.

### DIFF
--- a/games/The Witness.yaml
+++ b/games/The Witness.yaml
@@ -37,3 +37,12 @@ The Witness:
     20: 50
   puzzle_skip_amount: 
     10: 50
+triggers:
+  - option_name: shuffle_doors 
+    option_category: The Witness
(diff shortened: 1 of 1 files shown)


games/The Witness.yaml
```diff
@@ -1,8 +1,48 @@
...
+    10: 15
+    11: 5 
+  early_secret_area: 
+    false: 50
+    true: 50
+  trap_percentage: 
+    20: 50
+  puzzle_skip_amount: 
+    10: 50
+triggers:
```
ThePhar, line 40: This needs to be at the same level as all the other options or it will fail to generate a yaml.

ThePhar (changes requested):

ScootyPuffJr1:
Triggers look good

PR #37, opened 2022-10-01
SC2 - Difficulty changed to bell curve, more random options, new rele…

Magnemania:
…gate no build option

Game Difficulty:  Client can now alter game difficulty regardless of seed difficulty.  This option was originally added to the client at user request due to the lack of harder difficulty seeds in the previous async.  Now a bell curve centered around Normal difficulty.

Mission Order:  Weights reversed to favor mission shuffle over vanilla for increased randomization.

Shuffle Protoss:  Weighted in favor of shuffling for increased randomization.

Relegate No Build:  New option that moves no-build missions into endgame.  These missions have no logical requirements, so moving them to endgame reduces the likelihood of a player being capable of obtaining a very large number of checks very early.  Weighted in favor of not relegating as option is not as overwhelmingly popular as pure mission randomization.

Tested by generating a seed with 100 yamls with these settings and looking through the spoiler log.

### DIFF
--- a/games/Starcraft 2 Wings of Liberty.yaml
+++ b/games/Starcraft 2 Wings of Liberty.yaml
@@ -1,9 +1,9 @@
 Starcraft 2 Wings of Liberty:
   game_difficulty:
-    casual: 2
-    normal: 3
-    hard: 2
-    brutal: 1
+    casual: 20
+    normal: 30
+    hard: 20
+    brutal: 10
   upgrade_bonus:
     ultra_capacitors: 50
     vanadium_plating: 50
@@ -20,7 +20,7 @@ Starcraft 2 Wings of Liberty:
     false: 10
     true: 40
   relegate_no_build:
-    false: 40
-    true: 10
+    false: 50
+    true: 50
   progression_balancing:
     0: 50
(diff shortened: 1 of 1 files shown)


TheT113:
I would be in favor of making relegate_no_build be 3 true 2 false; I think it's a more worthwhile option in an async to have on, but also understand some people wouldn't want it on, I guess.

I don't feel strongly about this, but I figure I'd still weigh in my dissent in case  others feel this way.

games/Starcraft 2 Wings of Liberty.yaml
```diff
@@ -1,7 +1,9 @@
 Starcraft 2 Wings of Liberty:
   game_difficulty:
-    casual: 50
-    normal: 50
+    casual: 2
+    normal: 3
+    hard: 2
+    brutal: 1
```
t3hf1gm3nt, line 6: id turn these into powers of 10 if anything just for consistency with the other weights (so 20:30:20:10)

Magnemania: Fixed.

t3hf1gm3nt:
so i resolved our conversation without actually looking at the code again...methinks you reverted it back to what you had? because i dont see either of the changes reflected in the final code

Magnemania:
@t3hf1gm3nt Whoops, sorry about that, haven't amended a pull request before and botched it.  Should be fine now.

PR #38, opened 2022-10-09
Add Pokémon Red and Blue

Alchav:
Putting this here now for when/if Pokémon Red and Blue are added to AP

### DIFF
--- a/games/Pokemon Red and Blue.yaml
+++ b/games/Pokemon Red and Blue.yaml
@@ -7,7 +7,7 @@ Pokemon Red and Blue:
     items: 50
     minimal: 0
   game_version: # Select Red or Blue version.
-    red: 0
+    red: 50
     blue: 50
     random: 0
   elite_four_condition: # Number of badges required to challenge the Elite Four once the Indigo Plateau has been reached.
@@ -44,10 +44,10 @@ Pokemon Red and Blue:
     #    Scientist after reviving this number of fossils.
     # you can add additional values between minimum and maximum
     0: 0 # minimum value
-    3: 50 # maximum value
+    3: 0 # maximum value
     random: 0
     random-low: 0
-    random-high: 0
+    random-high: 50
   badgesanity: # Shuffle gym badges into the general item pool. If turned off, badges will be shuffled across the 8 gyms.
     false: 25
     true: 50
@@ -279,16 +279,33 @@ Pokemon Red and Blue:
     random-high: 0
     3000: 50
     random-range-2500-10000: 25
-  trainer_name: RED
-  rival_name: BLUE
   triggers:
+    - option_category: Pokemon Red and Blue
+      option_name: game_version
+      option_result: red
+      options:
+        Pokemon Red and Blue:
+          rival_name:
+            BLUE: 5
+            GARY: 3
+            JOHN: 1
+          trainer_name:
+            RED: 5
+            ASH: 3
+            JACK: 1
     - option_category: Pokemon Red and Blue
       option_name: game_version
       option_result: blue
       options:
         Pokemon Red and Blue:
-          trainer_name: BLUE
-          rival_name: RED
+          trainer_name:
+            BLUE: 5
+            GARY: 3
+            JOHN: 1
+          rival_name:
+            RED: 5
+            ASH: 3
+            JACK: 1
     - option_category: Pokemon Red and Blue
       option_name: randomize_pokemon_types
       option_result: vanilla
@@ -307,4 +324,4 @@ Pokemon Red and Blue:
             vanilla: 50
             prefer_types: 50
             completely_random: 5
-  
+
--- a/games/Pokemon Red and Blue.yaml
+++ b/games/Pokemon Red and Blue.yaml
@@ -1,0 +1,310 @@
+Pokemon Red and Blue:
+  accessibility: # Set rules for reachability of your items/locations.
+    #    Locations: ensure everything can be reached and acquired.
+    #    Items: ensure all logically relevant items can be acquired.
+    #    Minimal: ensure what is needed to reach your goal can be acquired.
+    locations: 0
+    items: 50
+    minimal: 0
+  game_version: # Select Red or Blue version.
+    red: 0
+    blue: 50
+    random: 0
+  elite_four_condition: # Number of badges required to challenge the Elite Four once the Indigo Plateau has been reached.
+    #    Your rival will reveal the amount needed on the first Route 22 battle (after turning in Oak's Parcel).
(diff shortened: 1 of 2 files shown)


TheT113:
I'm curious about the logic of having elite_four_condition be random-low. Async games go longer, so random-high or even like, random-range-4-8 would make the most sense to me. random-low combined with viridian_gym_condition random-high also makes viridian gym a rarity?

Similarly for Oak's Aides, they seem really low. It's an async, we have time to go catch a bunch of Pokemon!

exp_modifier: half seems cruel. I think "normal" is slow enough; if people want to challenge themselves, they would be submitting a custom.

randomize_wild_pokemon and etc, I think should have a higher chance of vanilla. There is only 1/3 a chance of vanilla, when maybe 50% would be better. similar for randomize_pokemon_stats, randomize_trainer_parties, etc.

randomize_pokemon_types randomize_type_matchup_types, randomize_type_matchup_type_effectiveness seem extremely spicy to me and also be a lower chance than vanilla.

I could also see adding a trigger that if pokemon are rolled vanilla (maybe randomize_wild_pokemon), that TMs should also be vanilla? But that's not one I'm pushy one.

Alchav:
> I'm curious about the logic of having elite_four_condition be random-low. Async games go longer, so random-high or even like, random-range-4-8 would make the most sense to me. random-low combined with viridian_gym_condition random-high also makes viridian gym a rarity?
> 
> Similarly for Oak's Aides, they seem really low. It's an async, we have time to go catch a bunch of Pokemon!
> 
> exp_modifier: half seems cruel. I think "normal" is slow enough; if people want to challenge themselves, they would be submitting a custom.
> 
> randomize_wild_pokemon and etc, I think should have a higher chance of vanilla. There is only 1/3 a chance of vanilla, when maybe 50% would be better. similar for randomize_pokemon_stats, randomize_trainer_parties, etc.
> 
> randomize_pokemon_types randomize_type_matchup_types, randomize_type_matchup_type_effectiveness seem extremely spicy to me and also be a lower chance than vanilla.
> 
> I could also see adding a trigger that if pokemon are rolled vanilla (maybe randomize_wild_pokemon), that TMs should also be vanilla? But that's not one I'm pushy one. (EDIT: I realized I misread the trigger at the end, which is instead based on if types are randomized... I guess that does make more sense in the long run. It's a little bit weird to have, types randomized + vanilla movesets be a possibility as well, honestly...)

Elite four condition is random high, victory road condition is random low. IIRC these mirror the crystals for GT and crystals for Ganon settings for the lttp filler

TheT113:
Ah, my bad. Misread those ones. While those are ok, range tends to be a little more preferable just to avoid the extremely low chance of rolling a 1 on elite 4 condition.

games/Pokemon Red and Blue.yaml
```diff
@@ -0,0 +1,327 @@
...
+    0: 0 # minimum value
+    255: 0 # maximum value
+    random: 0
+    random-low: 0
+    random-high: 0
+    half: 25
+    normal: 75
+    double: 75
+    triple: 50
+    quadruple: 25
```
t3hf1gm3nt, line 141: half seems too mean, but triple and quadruple seems too much. i'd probably do either 50/50 or 75/25 for normal/double and 0 out the rest

games/Pokemon Red and Blue.yaml
```diff
@@ -0,0 +1,327 @@
...
+      options:
+        Pokemon Red and Blue:
+          trainer_name:
+            BLUE: 5
+            GARY: 3
+            JOHN: 1
+          rival_name:
+            RED: 5
+            ASH: 3
+            JACK: 1
```
t3hf1gm3nt, line 308: is there a reason this needs to be in triggers? couldn't this just be in the base yaml as one option to save some lines?

Alchav: I don't know how else to achieve this effect

t3hf1gm3nt: OH i was misreading exactly how this was set up. i thought they were the exact same for each version, but now see that the options are flipped depending on what version is rolled. disregard

TheT113:
viridian_gym_condition is set to random-range-high-4-8, but the maximum is 7.

Alchav:
> viridian_gym_condition is set to random-range-high-4-8, but the maximum is 7.
> 
> How does the exp_modifier work with normal numbers? Like what is "normal" and "double" equivalent to

I knew that, I was just seeing if you knew 👀

Alchav:
> How does the exp_modifier work with normal numbers? Like what is "normal" and "double" equivalent to

16 is normal, 32 is double

TheT113:
Since the Oak's Aide route 15 check is missing it's logic rule, it'd probably be good to toss it in excluded locations for now, unless you're adamant a new AP bugfix release will come out before the next async.

Alchav:
> Since the Oak's Aide route 15 check is missing it's logic rule, it'd probably be good to toss it in excluded locations for now, unless you're adamant a new AP bugfix release will come out before the next async.

It sounds like there another AP release will occur in December, so I would guess this would be fixed before the next async

Alchav:
Converted to draft for now as, if I get my way, there will be new options by 0.3.7, which I'm guessing will be out before the next async

Alchav:
I have updated to add the new 0.3.7 options. I've also increased the chance of Pokémon Red and Blue games to be chosen due to the high popularity of this game - https://archipelago.gg/stats - however this entire list of weights seems to be quite random and a discussion about all the games' weights should probably be had.

Joethepic (commented):
I don't think require item finder should be random personally. I also find it very strange how moves can't be vanilla but trianers and wild can

t3hf1gm3nt:
my only comment on the update is that all the game weights will be re-evaluated once all the new games are in. i know pokemon is popular and we will try to balance the weights accordingly once we do so.

Alchav:
> I don't think to require item finder should be random personally I think it should always be true at the very least always true when hidden items are on. I also find it very strange how moves can't be vanilla but trainers and wild can.

```
  triggers:
    - option_category: Pokemon Red and Blue
      option_name: randomize_pokemon_types
      option_result: vanilla
      options:
        Pokemon Red and Blue:
          randomize_pokemon_movesets:
            vanilla: 50
            prefer_types: 50
            completely_random: 15
          tm_compatibility:
            vanilla: 50
            prefer_types: 50
            completely_random: 5
            full_compatibility: 25
          hm_compatibility:
            vanilla: 50
            prefer_types: 50
            completely_random: 5
```

Joethepic:
Ah  makes sense must of missed that

PR #39, opened 2022-10-10
meta support

Berserker66:


alwaysintreble:
i barely understand the original code that Phar wrote but I at least understand the changes being made here and lgtm. Will likely need to revisit meta resolution in core so we can use this for OptionLists, OptionSets and OptionDicts.

ThePhar:
Let me pull it and give it a whirl.

GenerateMystery.py
```diff
@@ -62,29 +63,42 @@ def main():
...
+            with open(meta_path) as meta_file:
+                meta_settings: dict = yaml.unsafe_load(meta_file)
+            if len(meta_settings.items()) > 1:
+                raise ValueError(f"Found {len(meta_settings)} top-level keys in `{meta_path}`")
+            if game not in meta_settings:
+                raise ValueError(f"Could not find `{game}` top-level key in `{meta_path}`")
+            meta.update(meta_settings)
 
+    with open("output/mystery.yaml", "w+") as file:
         yaml.dump(mystery, file)
```
ThePhar, line 96:
```suggestion
        yaml.dump(dict(mystery), file)
```

Looks like this now outputs the following info in the mystery.yaml file:

```
!!python/object/new:__main__.MysterySettings
dictitems:
  A Link to the Past:
```

which breaks generation (or at least fails the check on the website). Casting to dict seems to prevent that meta data about the class from getting written. Meta looks good though.

ThePhar (changes requested):

GenerateMystery.py
ThePhar: Probably because of the switch from using `mystery = {**mystery, **game_settings` to `mystery.update(game_settings`.

PR #40, opened 2022-11-05
Add new AreaRandomizer option to Super Mario 64 filler

Alchav:


Joethepic:
> I almost feel like courses_and_secrets should be dropped until the issue with TTC being at a Secrets entrance is fixed? It's a low chance to occur but it's still such an overly frustrating thing that people will complain about, that the chance might as well be removed.

I disagree with this just because having them fixed up is way more interesting while I’ve had ttc on secret courses and it can be frustrating it is possible someone else could do it for them worst case

t3hf1gm3nt:
It is a low enough chance for TTC to be in a place with no hands that I dont think its a big deal, especially with it being an even lower chance of courses_and_secrets even being selected. it makes red coins tedious but still possible if you cant manipulate the time. 100 coins though...unsure about, but theres a low chance of 100 coins being required anyway. I'll probably end up merging this as is but will wait a bit to get more opinions on the matter

Alchav:
I'm only trying to add the new option that I created to the filler yaml. I don't want to go making changes to existing weights, that should be done and discussed separately if desired.

t3hf1gm3nt:
fair enough. weight for the new option looks good to me so i'll merge this.

PR #41, opened 2022-11-13
Factorio: update factorio yaml with 0.3.6 options and some more variations

alwaysintreble:


games/Factorio.yaml
```diff
@@ -1,4 +1,12 @@
 Factorio:
+  start_hints:
+    - Progressive Science
```
t3hf1gm3nt, line 3: adding it as a start hint seems weird to me. if we are just wanting to let the fillers be able to more easily be able to do their first science, i'd much rather one science pack be added to start_inventory instead

ScootyPuffJr1: We chatted near the end of the last async about adding 1 start hint for a few games where early BK could be a problem.

ScootyPuffJr1:
Need to remove that start hint for progressive science, iirc.

PR #42, opened 2022-11-21
Update Timespinner

t3hf1gm3nt:
- Make Gears local items
- Update FacebookMode to new name EyeSpy
- Change StartWithMeyef to true (you get Meyef as soon as you die for the first time anyway)
- Update Inverted and GyreArchives to use random chance
- Change Cantoran from always false to random (with the new collect functionality of Timespinner, getting stuck behind Cantoran is no longer an issue because a player can just start a new save)
- Add BossRando with a random chance
- Update DamageRando option from false to off
- Add ShopFill set to default
- Add ShopWarpShards set to true
- Add LootPool set to randomized
- Add DropRateCategory set to randomized
- Add LootTierDistro set to full_random
- Add ShowBestiary set to true
- Add ShowDrops set to true
- Add EnterSandman with a quarter chance of being true

### DIFF
--- a/games/Timespinner.yaml
+++ b/games/Timespinner.yaml
@@ -7,19 +7,11 @@ Timespinner:
   StartWithMeyef: true
   QuickSeed: false
   SpecificKeycards: true
-  Inverted:
-    false: 50
-    true: 50
-  GyreArchives:
-    false: 50
-    true: 50
-  Cantoran:
-    false: 50
-    true: 50
+  Inverted: random
+  GyreArchives: random
+  Cantoran: random
   LoreChecks: true
-  BossRando:
-    false: 50
-    true: 50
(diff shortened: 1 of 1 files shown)


Joethepic (commented):
Other then maybe making gears not local forced because they do unlock a few checks if enter sandman is off however it's not a lot of checks so if you feel confident on leaving them local I think it's good.

t3hf1gm3nt:
if you dont have enter sandman you're finishing the game by time you get all the gears anyway, so thats moot
there was discussion in the big changes discord thread that making gears local for fillers was a thing people would like to at least try, to reduce the amount of time sitting in BK with an otherwise full cleared game besides pyramid checks

Joethepic (approved):

PR #44, opened 2022-11-25
add Oc2

Joethepic:


ScootyPuffJr1:
This will not work with the current process. These files do not need a name or game line. Also, please clean up as many unnecessary rows as you can. It seems like 80% of this can be cut out.

Joethepic:
Should work now also removed unnecessary lines.

PR #45, opened 2022-11-27
Add SMW

TheT113:
My suggestions for Super Mario World.

Going with only Bowser goal since I assume Yoshi Egg Hunt would be discouraged/expected local.

### DIFF
--- a/games/Super Mario World.yaml
+++ b/games/Super Mario World.yaml
@@ -1,3 +1,4 @@
+Super Mario World:
   goal: bowser
   bosses_required: random-range-5-10
   dragon_coin_checks:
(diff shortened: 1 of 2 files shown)


t3hf1gm3nt:
i think for this first try with SMW fillers in the async, no yoshi egg hunt will be fine. will probably want to add it later at some point but thats for future us to figure out a good balance for. otherwise looks good but will leave it up a little longer to see if anyone else has comments

t3hf1gm3nt:
the meta yaml is gonna get re-evaluated after im done merging in any changes and before we actually roll. gonna confer with Scooty on what the final weights should be. so it just existing in the weight pool is good enough for me
i think a 1/3rd chance of dragon coins on is fine. some of those checks can be a bit difficult and a smaller chance of it being on mirrors KONGsanity

Joethepic (approved):

TheT113:
Yeah; I am 100% fine with doing whatever to the meta weights. They definitely need an overhaul and I'm not at all equipped to consider that.

PR #46, opened 2022-12-26
Update SA2 to 0.3.7.

TheT113:
My rationale behind these choices:

Chaos Emeralds can be a bit of a slog and Triforce Hunt-adjacent (more so than Emblems are), so let's leave it out.

Chao Keys and Pipes are doable (and there's a good guide out), but let's keep the chance of them lower. Beetles in general are very easy, but keeping the same random weights for consistency.

For level types, a little bit of variety is good. Keeping Chao missions (Mission 3) in the rotation ensures the Mystic Melodies have a chance to be required, and otherwise the missions are not hard. Ring Missions (Mission 2) are typically very easy. Speed/Mech Mission 4s (Timer) are generally interesting and worth having a chance to show up.

Speed Missions randomly have 2-4 missions each because Speed missions are generally the most fun and interesting. If it rolls 2 or 3, it guarantees there will be variety in the mission type seen; if it rolls 4, at least they will be in random order.

All other missions are 1 or 2, with some variety to allow different mission types to be experienced.

Kart Mission 2 are extremely easy and if Mission Count is rolled 1, the chance that people will only have to do Rings and not the normal run would be a relief.

### DIFF
--- a/games/Sonic Adventure 2 Battle.yaml
+++ b/games/Sonic Adventure 2 Battle.yaml
@@ -50,12 +50,12 @@ Sonic Adventure 2 Battle:
   speed_mission_3: true
   speed_mission_4: true
   speed_mission_5: false
-  mech_mission_count: random-range-1-2
+  mech_mission_count: random-range-1-3
   mech_mission_2: true
   mech_mission_3: true
   mech_mission_4: true
   mech_mission_5: false
-  hunt_mission_count: random-range-1-2
+  hunt_mission_count: random-range-1-3
   hunt_mission_2: true
   hunt_mission_3: true
   hunt_mission_4: false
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
i still have yet to play SA2 so i can't really comment on these changes. will leave this up and try to prod some people to weigh in. if theres no major objections will probably just end up merging

Joethepic (commented):
For a async I think the mission counts are kinda low for specifically hunt and maybe mech. Though I'm not great at sonic yet but with more time more missions make sense to me and their is already 3 missions in rotation I think random-range-1-3 would work better for hunt.

TheT113:
These are beginner worlds. People who want a want a longer, harder world will be submitting a custom.

The length of a world in an async is immaterial. As long as it isn't possible to go mode it extremely early, the length doesn't matter. The gating does.

TheT113:
Thought about it again. I'll increase the chance to 3 for Mech and Hunting.

Joethepic (approved):

PR #47, opened 2022-12-26
SC2 - New options and campaign types

Magnemania:
Handles the relegate_no_build -> shuffle_no_build renaming, removes the unrandomized mission order from the campaign types, adds the two longest new mission orders, and locks early_unit + required_tactics + units_always_have_upgrades to beginner/async-friendly settings.

### DIFF
--- a/games/Starcraft 2 Wings of Liberty.yaml
+++ b/games/Starcraft 2 Wings of Liberty.yaml
@@ -8,7 +8,8 @@ Starcraft 2 Wings of Liberty:
   bunker_upgrade: random
   all_in_map: random
   mission_order:
-    vanilla_shuffled: 40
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
thinking on it: feels weird to not have any chance for vanilla mission order. i'd leave a small chance for it for those really new players who wanna try it out

Magnemania:
> thinking on it: feels weird to not have any chance for vanilla mission order. i'd leave a small chance for it for those really new players who wanna try it out

I think mission order shuffle is a bit less newbie-unfriendly than it is for most level randos, since it's guaranteed to put easier missions early, but I can take 20% away from the other settings for vanilla.

PR #48, opened 2022-12-26
Update meta yaml to require 0.3.7.

TheT113:
This probably isn't important!


Berserker66:
Indeed, the meta.yaml doesn't contain any changes that are 0.3.7 only.

PR #49, opened 2022-12-27
Ocarina of Time 7.0 async settings

espeon65536:
This is a pretty extensive compression of the YAML. Of note:
- Mask of Truth is excluded now because it sucks
- There is a chance for local triforce hunt.
- Chest appearances and pot appearances match their contents.
- GTBK on LACS has been converted to GTBK granted on medallion/reward count (essentially the same though)
- Boss keys have a chance for keysanity. Small keys have a chance to be keyrings; if they are, they get shuffled into the general multiworld pool.
- There is a 10% chance for dungeon pot shuffle. This is sort of a test of how it goes, since this potentially adds a lot of locations into filler seeds. If it goes poorly, we can remove it in the future.

### DIFF
--- a/games/Ocarina of Time.yaml
+++ b/games/Ocarina of Time.yaml
@@ -1,210 +1,112 @@
 Ocarina of Time:
-  logic_rules: # Determine the logic used to place items.
-    glitchless: 1 # No glitches are required, but may require some minor tricks (can be set under logic_tricks)
-    glitched: 0 # Movement-oriented glitches are required. Entrance randomization of any kind is not compatible with glitched logic.
-    no_logic: 0 # All locations considered available with no items. May be impossible to beat.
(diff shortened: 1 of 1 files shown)


Joethepic (commented):
What's the reason on doing always door of time i personally think it should be random in some capasity for those who want smaller sphere 1. Is there also plans on beehives or is that falling on the other crazy shuffle rules.

t3hf1gm3nt:
closed DoT i feel is one of those things that if you want it on, submit a custom. fillers are meant to be more accessible and some checks that can be done in both ages are easier in one age compared to the other.
beehives i have no opinion one way or the other and leave that up to espeon's judgement (or more feedback from other users). i dont think it was left out because of possible gen issues (dont think they caused any last i looked)

PR #50, opened 2023-01-13
Changing game weights

ScootyPuffJr1:
This was my note from the start of the October async: "About 24 hours after start and no Raft, SA2B, or Subnautica seeds claimed. SoE and V6 both have a ton open. SM64 full, SC2, StS, OOT, LttP, HK, MC mostly full." New games were mostly guesses. There are a lot of people with the Pokemon and SMW roles. In order to find a few percents for HK, OOT, and SM64, I took some from TS and RL. I'd be fine if we lowered the LttP number, but those fillers tend to go fast.

### DIFF
--- a/games/__meta__.yaml
+++ b/games/__meta__.yaml
@@ -6,23 +6,27 @@ requires:
 # Define your game here!
 game:
   A Link to the Past: 80
-  Factorio: 15
-  Minecraft: 30
-  Slay the Spire: 15
-  Risk of Rain 2: 15
-  Subnautica: 5
-  Ocarina of Time: 25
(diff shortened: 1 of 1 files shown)


Joethepic (approved):
I still slightly question if RL should be down by 10 and not something closer to 5 but I see why.

t3hf1gm3nt:
They look fine to me. Going to leave this open for just a bit longer in case we get any other game submissions last minute (since we are still missing Hylics, Lufia, and Zillion)

Joethepic:
Of course #45 #44 need to be merged before this and #38

PR #52, opened 2023-01-13
Add new options to Subnautica filler

SnowOfAllTrades:
Adds the new yaml options "Early Seaglide" and "Swim Rule" to the filler yaml. Weights are what I came up with on the spot, may need to be adjusted. Swim Rule has no weights for Hard options since they can require death runs and I therefore consider them too hard for a filler, Easy has higher chances than Normal to ensure beginners picking up a seed have a better time.

### DIFF
--- a/games/Subnautica.yaml
+++ b/games/Subnautica.yaml
@@ -14,3 +14,11 @@ Subnautica:
     stasis: 5
     containment: 2
     either: 2
+  early_seaglide:
+    true: 2
+    false: 1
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
maybe bump the weight of early seaglide up to 3, to make it only a 25% chance of non-early seaglide. looks good to me otherwise

SnowOfAllTrades:
done

PR #54, opened 2023-01-14
[OC2] Update yaml

toasterparty:


Joethepic:
Couple things i turned on kevin levels to make the seeds longer same as shorter durations because in an async how fast you send checks doesn't matter. Also a 50 on all options is the same as random and they perfered into small as possible

PR #56, opened 2023-01-21
Updating Factorio and Raft so they can roll without failing

ScootyPuffJr1:


TheT113:
(This post is just an anecdote, not anything related to changes, feel free to not read)

lmao; I was test rolling for fun and wasn't sure why Factorio failed sometime, and I just assumed I had done something wrong on my end; I didn't bring it up because I assumed "well, it's not my place to be test rolling the yaml anyway, and if it's working for Berserker then it's clearly my end, so whatever."

So uh, sorry! Glad it's figured out though!

PR #57, opened 2023-02-04
Update Factorio.yaml

Berserker66:
Not tested, just written up in github's editor.

Removed traps, they're bad for async.
Revamped tech costs to be reasonable
added recipe time, tech cost distribution and ramping tech cost

Personally I'd also recommend removing the peaceful mode starts. If someone wants to play with peaceful they can use the server console to change it.

### DIFF
--- a/games/Factorio.yaml
+++ b/games/Factorio.yaml
@@ -19,20 +19,19 @@ Factorio:
     trees: 1
     choices: 1
   min_tech_cost:
-    5: 50
-    random: 20
-    random-low: 20
-    random-high: 20
+    1: 50
   max_tech_cost:
-    500: 50
-    random: 20
-    random-low: 20
-    random-high: 20
+    random-range-low-100-10000: 20
+  tech_cost_distribution:
+    even: 10
+    low: 10
+    middle: 10
+    high: 10
+  ramping_tech_costs:
+    true: 25
+    false: 50
   tech_cost_mix:
-    70: 50
     random: 20
-    random-low: 20
-    random-high: 20
   silo:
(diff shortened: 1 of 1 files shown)


alwaysintreble:
While I agree with you on the peaceful mode stuff, enough people complain about it that I felt it was only fair to give them the option to start that way without having to enter in a command.

t3hf1gm3nt:
> Personally I'd also recommend removing the peaceful mode starts. If someone wants to play with peaceful they can use the server console to change it.

does the reverse also apply? using console to switch it to non-peaceful? if so: id rather it be all peaceful starts and those who want to deal with biters can turn it on themselves

t3hf1gm3nt:
this probably needs updated with the new factorio things, doesnt it?

Berserker66:
Probably does, good luck, please don't mess it up again.

Berserker66:
> 



> > Personally I'd also recommend removing the peaceful mode starts. If someone wants to play with peaceful they can use the server console to change it.
> 
> does the reverse also apply? using console to switch it to non-peaceful? if so: id rather it be all peaceful starts and those who want to deal with biters can turn it on themselves

No, the thing most people use to turn it peaceful removes the Biters from the map, which is irreversable.

t3hf1gm3nt:
> Probably does, good luck, please don't mess it up again.

no i meant aren't there new options? i dunno jack about factorio
so was wondering if you wanted to update this with said new options before i do anything with this

Berserker66:
This PR's purpose is to unfuck Factorio.

PR #58, opened 2023-02-10
Use more acessible settings for new players

Ziktofel:
How I can test generating the game?

### DIFF
--- a/games/Starcraft 2 Wings of Liberty.yaml
+++ b/games/Starcraft 2 Wings of Liberty.yaml
@@ -1,19 +1,39 @@
 Starcraft 2 Wings of Liberty:
   game_difficulty:
-    casual: 20
-    normal: 30
+    casual: 25
+    normal: 35
     hard: 20
-    brutal: 10
+    brutal: 0
   upgrade_bonus: random
   bunker_upgrade: random
   all_in_map: random
   mission_order:
-    vanilla: 10
+    vanilla: 5
     vanilla_shuffled: 30
-    grid: 30
-    mini_campaign: 30
-  shuffle_protoss: random
+    mini_campaign: 0 # Significantly harder All in
+    grid: 0 # Significantly harder All in
+    mini_grid: 20 # No All in
+    blitz: 25 # No All in
+    gauntlet: 20 # No All in
+  shuffle_protoss: false # Set to random for vanilla or vanilla_shuffled by a trigger
   shuffle_no_build: random
   early_unit: true
   required_tactics: standard
   units_always_have_upgrades: true
+  triggers:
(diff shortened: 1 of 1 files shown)


Ziktofel:
In the current async runs all unclaimed fillers so far are the shorts with All in - In these settings All in is significantly harder than other missions

The purpose of this change is to offer fillers accessible to new players

Magnemania:
Short seeds are difficult for new players because they place more difficult missions earlier in the seed and provide fewer extra items beyond the bare minimum logic required to complete the seed.

In an async, the tiny item pool is especially problematic because players receive only a tiny fraction of their items at the beginning; it's very possible for a short seed's only anti-air unit to be stuck at the end of a long game, and the algorithm is unlikely to care, as blocking an entire short seed only blocks ~30 locations.

It is also fundamentally against the spirit of the async to play games on settings intended for <2 hour runs.

Also, **shuffle_protoss** does nothing on vanilla; that trigger is unnecessary.

The only part of this that I agree with is the difficulty rebalancing.  Experienced players should be aware of the /difficulty command, so biasing the seeds even more heavily towards lower difficulties should be fine.  However, I'm somewhat opposed to making Casual an extremely common option, as logic barely applies on Normal and never applies on Casual.

Ziktofel:
What about the increased difficulty of all in in shorter campaigns for new players?

Ziktofel:
Is there a way to reduce the time commitment (from full campaign) without increasing the difficulty of All-in?

Ziktofel:
@Magnemania I think the difficulty problem of shorts applies mainly to gauntlet because you're forced to play certain mission while in Blitz and minigrid you have a way around. Gauntlet could be dropped because of that

I still think that a shorter run that's not difficult has its place in async. In current situation the Blitz has 12 missions and it's the longest without all in. I'm open myself for large grid without `All in` or using somewhat nerfed version of `All in` in shorts

I still believe that Blitz could do a perfect job for a side-game

Magnemania:
Blitz can be completed in under two hours and requires even less logic than the other short seeds; that is entirely against the point of the async.

Ziktofel:
You can either complete fast when your goal is for example Media Blitz (there's cheese) or be in a hard BK for a long time. All depends on what stuff you have early on and cheesable missions

Ziktofel:
Tried the new All-out and with a good inventory it's a great ending for shorter campaigns (large grid, etc) :)

t3hf1gm3nt:
im actually going to close this. i agree with mag's concerns that a lot of this is against what the async is about. if people want shorter campaigns, they can submit customs. and the difficulty change is honestly not that big a deal with the change difficulty command existing.

PR #61, opened 2023-03-31
Adding Links Awakening

ScootyPuffJr1:


games/Links Awakening DX
```diff
@@ -0,0 +1,42 @@
...
+    normal: 5
+    hard: 2
+  shuffle_nightmare_keys:
+    original_dungeon: 5
+    own_world: 3
+    any_world: 2
+  shuffle_small_keys:
+    original_dungeon: 5
+    own_world: 3
+    any_world: 2
```
t3hf1gm3nt, line 22: im conflicted on the any_world option. as we keep keys local in LTTP but not in OOT, so unsure what the precedent should be. i could go either way on this decision but wanted to bring this up

games/Links Awakening DX
```diff
@@ -0,0 +1,42 @@
...
+    false: 3
+    true: 2
+  goal:
+    instruments: 5
+    seashells: 2
+  instrument_count:
+    random-high: 1
+  trendy_game:
+    normal: 5
+    hard: 2
```
t3hf1gm3nt, line 14: part of me wants a chance of easy, but im biased as thats what i personally run.

games/Links Awakening DX
ScootyPuffJr1: I removed the any world chances. I personally wouldn't want to play them so I figured the LttP way was better.

games/Links Awakening DX
ScootyPuffJr1: Sure, why not

t3hf1gm3nt:
Actually, one minor issue
The file isn't saved as a yaml file. Unsure how that happened but it will cause issues if its not a yaml file

t3hf1gm3nt:
Alright looks good now. Letting it simmer a few days before merging to see if anyone else has comments/suggestions

PR #62, opened 2023-04-05
Add Adventure

TheT113:
Not much to talk about here. Having the difficulty switches in the pool just adds some variety and interesting items, since there aren't a ton of items here to begin with. Because it's an async, even if the dragon speeds roll higher than vanilla... you have unlimited tries and infinite time to get around them, so it's not unmanageable; and there will be items in the pool to reduce them to their vanilla speeds if needed. Everything else doesn't change too much, just some variety.

I think 10 is a good weight, it's on par with Meritious and Evermore I feel. I almost originally put 5 until I saw how much higher everything else is...

### DIFF
--- a/games/Adventure.yaml
+++ b/games/Adventure.yaml
@@ -1,0 +1,23 @@
+Adventure:
+  dragon_slay_check: true
+  death_link: false
+  bat_logic:
+    cannot_break: 1
+    can_break: 2
+  freeincarnate_max: 17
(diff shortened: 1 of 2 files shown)


t3hf1gm3nt:
I know basically nothing about this game, but going by the option descriptions and the fact you added suggestions straight from the dev, I will probably approve this as is. Just gonna let it simmer a bit to see if there are any other comments/suggestions

PR #63, opened 2023-04-05
Update SA2 to AP 0.4.0

TheT113:
Tossing in the chance for Omosanity and Kart Racing at the same chance as everything else. Kart Racing is limited to mini as full can be a slog.


Joethepic (changes requested):
I would think ring loss might want to be put as a 2 modern /1 classic due to being slightly easier

t3hf1gm3nt:
Seems fine. From what little i know, I kinda agree with squidy about the ring behavior, so would like either a change or a good justification to leave as is first before I merge this

PR #64, opened 2023-04-05
Update SMW to 0.4.0

TheT113:
Nothing surprising here. I haven't had a chance to try bowser_castle_rooms: gauntlet, so I'm leaving it out as I assume it'd be too daunting. Two room should be manageable but still keeping vanilla as the higher weight. Adding exclude_special_zone with a higher weight to help lead to slightly easier worlds on average. Boss shuffle is really harmless so having it 2/3 chance isn't a risk.

Deleted some stuff that was set to default anyway, and changed the trigger for no reason at all really.

### DIFF
--- a/games/Super Mario World.yaml
+++ b/games/Super Mario World.yaml
@@ -5,23 +5,30 @@ Super Mario World:
     false: 2
     true: 1
   bowser_castle_doors: vanilla
+  bowser_castle_rooms:
+    vanilla: 2
+    random_two_room: 1
   level_shuffle:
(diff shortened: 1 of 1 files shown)


Joethepic (approved):
I see no issues with these changes but I haven't played the new option for bowser castle rooms yet. Might be worth someone else to check as well because I haven't played a lot of smw.

t3hf1gm3nt:
Generally approve of this. Gonna let it simmer a few days to see if there are any other comments first.

PR #65, opened 2023-04-05
Create Stardew Valley Filler

SnowOfAllTrades:
My first draft for a Stardew Valley filler. No Fishsanity, early_progressive on QoL options.

### DIFF
--- a/games/Stardew Valley
+++ b/games/Stardew Valley
@@ -1,0 +1,25 @@
+Stardew Valley:
+  accessibility: items
+  starting_money: 2000
+  resource_pack_multiplier: 100
+  bundle_randomization: thematic
+  bundle_price: normal
+  entrance_randomization: disabled
+  backpack_progression: early_progressive
+  tool_progression: progressive
+  skill_progression: progressive
+  building_progression: progressive_early_shipping_bin
+  elevator_progression: progressive
+  arcade_machine_locations: disabled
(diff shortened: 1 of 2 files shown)


TheT113:
The lack of variety here is... stark. I'd say, for fun, add more posibilities with starting money with a random-range. Have a chance of vanilla or thematic for bundle_randomization, and a chance for bundle_price: cheap to make things easier for people picking up a filler who might not be as experienced. I'm torn on goals; CC is really the best option, but maybe a small chance of one of the others to add some variety? Bottom of the Mines might be a good one, since it will offer some easier worlds for new players.

I still don't understand why someone would *want* debris_multiplier, so forcing it to half seems like an odd choice to me.

And turning gifting off seems like a huge mistake. It should be left open as an option to help struggling players. Also not sure why gift_tax is there if you set it off, but... 20 seems fine.

TheT113:
Also, it's not a .yaml file. Not sure if that would break things or not.

SnowOfAllTrades:
I agree on most accounts, except gifting. Gifting really breaks logic, since it basically means as soon as one Stardew world unlocks an item, ALL worlds with gifting have access to that item. You could fill out your entire CC (except for Gold Star Crops) just with gifts from other players and don't even need to get the logically required items for it, which I think is really against the spirit of the async.
Made some changes.

PR #66, opened 2023-04-05
Update ror2 to 0.4.0

Joethepic:
adds explore mode as options and changes some other settings to make more simple and add more variety

### DIFF
--- a/games/Risk of Rain 2.yaml
+++ b/games/Risk of Rain 2.yaml
@@ -11,12 +11,9 @@ Risk of Rain 2:
   total_revivals: random-high
   start_with_revive: 'true'
   final_stage_death:
-    # Death on the final boss stage counts as a win.
     false: 50
     true: 100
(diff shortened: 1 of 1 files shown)


kindasneaki:
A few things that I think I'd change would be:

- Change explore chance to be a little bit higher. 75 to 25 or 100 to 25.
- Have Chest per stage be `random-range-high` to prevent getting a really bad roll and getting the minimum amount a chests per stage. Maybe something along the lines of 5-15.
- I would rather have final stage death always be false but this is for beginners so idk. Id at least make the weights even.
- Maybe change `item_pickup_step: random-range-0-3` to `item_pickup_step: random-range-low-0-3`

t3hf1gm3nt:
My knowledge is minimal, but from what I can tell these changes seem fine. Will let this simmer to see if there are any other comments/suggestions

PR #67, opened 2023-04-06
Added Stardew Filler Yaml

agilbert1412:


TheT113:
Wow, this one seems a little too wild to me.

Entrance Randomization stands out as an odd choice. I don't feel like a beginner-friendly filler should have entrance rando on at all. If it did, it'd have to be weighted really low, but at 4x the rate of disabled? Way, way too high.

Some of the _progression options, are such an integral part of the SDV rando that I don't even think vanilla should be an option, but at 1/10 chance it's probably fine as is if you super feel it's needed.

For goal, I definitely think community_center should be the highest chance, with the others much lower, some of them maybe not even at all. If people want the harder/wilder goals, they wouldn't be grabbing a filler.

I think it'd be worth cutting out the higher ends of gift_tax, or even not randomizing at all and keeping it closer to soft, because it feels needlessly confusing to a new player to not know much they'll have to pay when (sending? it's sending right?) gifts.

fwiw, a lot of the options seem more interested in creating a chaotic and weirdly randomized seed, rather than one effective for beginners.

I'd also recommend cleaning this up a lot; removing the local_items through item_links section since they're not being used, and remove any comments or unselected options. Less important but it's strange you have so many numbers divisible by 10 when you could just cut the 0 off them.

games/Stardew Valley.yaml
```diff
@@ -0,0 +1,178 @@
...
+    []
+  start_location_hints:
+    []
+  exclude_locations:
+    []
+  priority_locations:
+    []
+  item_links:
+    []
+
```
t3hf1gm3nt, line 18: you can remove all this since its not necessary. will make things look neater

games/Stardew Valley.yaml
```diff
@@ -0,0 +1,178 @@
...
+  bundle_randomization:
+    vanilla: 25
+    thematic: 50
+    shuffled: 25
+
+  bundle_price:
+    very_cheap: 25
+    cheap: 50
+    normal: 25
+    expensive: 25
```
t3hf1gm3nt, line 39: personal opinion: but i'd drop expensive down a bit and maybe add it to normal. probably just by 10 at the most

games/Stardew Valley.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+  entrance_randomization:
+    disabled: 20
+    pelican_town: 40
+    non_progression: 40
```
t3hf1gm3nt, line 44: i might be biased as i've hardly messed with ER, but i'd weight this more aggressively towards no ER. like a 70/20/10 or 60/30/10

games/Stardew Valley.yaml
```diff
@@ -0,0 +1,178 @@
...
+    8: 10
+    10: 10
+    12: 5
+
+  goal:
+    community_center: 30
+    grandpa_evaluation: 30
+    bottom_of_the_mines: 10
+    cryptic_note: 20
+    master_angler: 10
```
t3hf1gm3nt, line 103: i'd probably drop cryptic note a bit and add it to CC. by 5 or 10 i think

agilbert1412: Fixed

games/Stardew Valley.yaml
agilbert1412: I did weight it towards no ER, even though I don't really agree as I believe in its current state, the stardew ER is more comical than anything else and doesn't really affect difficulty.
I kept pelican town and non-progression with equal weights, as they are not really fundamentally different from each other, it's just like 6 extra entrances behaving the same way.

So I made it 60/20/20, I think you'll be okay with that?

games/Stardew Valley.yaml
agilbert1412: Changed it to 15/50/25/10. I also noticed that my original one totalled to 125, which was not intended (I like to keep them at 100 for readability.

Expensive is a bit lower like you asked, and I split the reduction between normal and cheap. I personally believe that Cheap is the "default" experience for AP, because it allows at least a little bit of freedom for all bundles, instead of forcing the player to get almost every item.

games/Stardew Valley.yaml
agilbert1412: Done

t3hf1gm3nt:
so after some discussion, we've decided to go with Snow's PR for this coming async. Considering this is the first async SDV will be present in, we wanted to go with the better chances of more beginner friendly options and some simpler settings. We would like to go with the settings you've proposed in this PR in future asyncs, but with a few more caveats:
1) Deathlink off. DL is VERY brutal in SDV and one death could be very discouraging to someone picking up a filler. 
2) Gifting off. Really messes with the logic and reduces the need for SDV players to receive some of their items, which is against the spirit of the async
3) Update with SDV 3.x.x options, as since it is merged into AP main now, will definitely be available for future asyncs
up to you whether you just want to rebase this PR or close it and start with a fresh PR

ScootyPuffJr1:
@agilbert1412 Are you willing to modify this to include any 0.4.2 options?

PR #69, opened 2023-04-08
Update Pokémon R/B options

Alchav:


games/Pokemon Red and Blue.yaml
```diff
@@ -88,9 +88,30 @@ Pokemon Red and Blue:
...
+    false: 3
+  all_pokemon_seen:
+    true: 1
+    false: 1
   trainersanity:
     true: 1
     false: 3
+  randomize_rock_tunnel:
+    true: 1
+    false: 1
```
t3hf1gm3nt, line 105: so could you explain a bit better what the warning is about with this particular option? like is it still possible for things to be unreachable in its current state? if so im not sure this should be an option for the filler yaml until these issues are resolved

t3hf1gm3nt:
also congrats on getting PR 69. nice

games/Pokemon Red and Blue.yaml
Alchav: In the sense that it's possible that I made a mistake that hasn't been yet discovered

t3hf1gm3nt: For safety's sake until you are absolutely certain there are no issues with this option, I'm gonna request this option always being off for the fillers.
Once that's done, everything else looks fine on this and I'll merge it

PR #71, opened 2023-04-08
Rogue Legacy: Tweaks to RL filler yaml.

ThePhar:
Make things a little slower, but not ridiculously hard.


ScootyPuffJr1:
Both the chests_per_zone sections have range-range- instead of random-range-

ThePhar:
Son of a gun.

PR #73, opened 2023-04-08
Update Timespinner

t3hf1gm3nt:
Add Dadpercent, RisingTides, and UnchainedKeys.
Leaving out TrapChance should default it to 10% which I feel is fine.


Joethepic (approved):
personally i think dad percent could be even considering it's just a different goal but it works.

PR #77, opened 2023-04-16
Changing Game weights, Adding SDV

ScootyPuffJr1:
Here is my comment near the start of the last async: Looking at unclaimed fillers, LttP, DS3, DKC3, OC2, pokemon, SoE, SMZ3, SC2, SM, TS, V6 are candidates to have lower weights with SC2 being the least filled. LttP makes sense because people like to have multiple of those and there are a lot.

### DIFF
--- a/games/__meta__.yaml
+++ b/games/__meta__.yaml
@@ -5,33 +5,34 @@ requires:
 
 # Define your game here!
 game:
-  A Link to the Past: 80
-  Adventure: 10
-  Dark Souls III: 15
-  Donkey Kong Country 3: 20
-  Factorio: 15
-  Hollow Knight: 35
-  Links Awakening DX: 20
-  Meritous: 10
-  Minecraft: 30
-  Ocarina of Time: 30
-  Overcooked! 2: 15
-  Pokemon Red and Blue: 30
-  Raft: 3
-  Risk of Rain 2: 15
-  Rogue Legacy: 30
-  SMZ3: 30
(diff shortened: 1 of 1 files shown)


TheT113:
I'm lukewarm on this. I'm not a massive fan of rebasing the entire thing, and I had to make a spreadsheet to see what even really changed. After looking at them, I think I'm fine with it and 100 is a much better LttP base than 500. Here's a comparison for anyone who really needs it, old on the left, new on the right, SDV excluded.

image

I think Adventure is a bit too low, and would rather see it equivalent to Evermore; but I'm not sure how much of that is bias talking. (Either way, it's still extremely low, I don't think the the extra .45% will make a difference). I could even see Meritous be lowered to be the same level as Adventure and Evermore.

I agree with making LoZ higher, 10 was too low for it on the old scale.

I feel like MC is a bit too high now, and would probably fit in with the other 40s. Are TS and HK really that super popular? I guess so, huh... they just seem really high. Similarly with SM64...

But with that said, the percentages are still so close to each other that it won't be radically different in the long run. I think this is probably still a good base and would personally approve it; the natural caveat being "we'll see how the fillers fill up this time and adjust for next", as usual.

So no issues here (except maybe Adventure, hehe.)

TheT113:
Since roles are no longer required to join rooms, that's not a good indicator of determining interest in a game. Looking at the last few chatters in that room, the majority do not have the Adventure role.

PR #81, opened 2023-05-28
Update StS yaml to use new character IDs

sylvia43:
From the community async, we had a couple seeds that should have been different characters defaulting to The Ironclad.

> async slot 889 claims to be silent, but is actually ironclad when you connect - can anyone else check this? it's definitely possible that my setup is wrong, but I haven't experienced anything like this before

> my seed (Player976) is also ironclad instead of silent?

Looked like the problem was the `character` format changing with a more recent update, this PR updates those to the new format from the current template YAML.

### DIFF
--- a/games/Slay the Spire.yaml
+++ b/games/Slay the Spire.yaml
@@ -1,9 +1,9 @@
 Slay the Spire:
-  character: # Pick What Character you wish to play with.
-    ironclad: 50
-    silent: 50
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
@KonoTyran just wanted confirmation: does this look correct?

KonoTyran:
Yes it does

PR #83, opened 2023-06-15
Add Noita

ScipioWright:
I talked over the weighting with the other maintainer, Heinermann, and he was fine with it.

Tried generating with it with higher Noita rates and other games several times, and it worked even with extreme conditions, so looks good to me.

### DIFF
--- a/games/Noita.yaml
+++ b/games/Noita.yaml
@@ -1,13 +1,13 @@
 Noita:
   bad_effects: 'true'
   victory_condition:
-    greed_ending: 20
+    greed_ending: 0
     pure_ending: 50
-    peaceful_ending: 0
+    peaceful_ending: 50
   path_option:
-    main_path: 30
+    main_path: 0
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
@ScipioWright Just checking with you if any of these options need to be updated to be up to date with AP version 0.4.2

ScipioWright:
Oh you're totally right, I need to make a minor update to this to add in some of the newer options. I'll do that later today, thank you for the reminder!

ScipioWright:
Converted to draft since I messed with it in-browser. Will test it once I get home and un-draft it then.

ScipioWright:
This is now ready, sorry for the delay!

ScootyPuffJr1:
I've verified that this runs and the game weight seems fine. I don't know enough about the game to comment on option weights, but we've chatted about them in Discord, so they are probably good to go.

PR #84, opened 2023-06-24
The Messenger: adjustments post 0.4.1

alwaysintreble:
makes the keys local and adds a chance for mega shards to be shuffled. shop shuffle is always enabled, when that gets merged, and the only new option being added with it is plando costs so no additional yaml updates necessary.

### DIFF
--- a/games/The Messenger.yaml
+++ b/games/The Messenger.yaml
@@ -4,11 +4,11 @@ The Messenger:
     - Keys
   goal:
     open_music_box: 80
-    power_seal_hunt: 10
+    power_seal_hunt: 20
   music_box:
     true: 100
-    false: 10
-  percent_seals_required: random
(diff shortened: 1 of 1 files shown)


games/The Messenger.yaml
```diff
@@ -9,3 +10,6 @@ The Messenger:
     false: 10
   percent_seals_required: random
```
ScootyPuffJr1, line 11: What do you think about forcing this to be at or above 70%?

t3hf1gm3nt:
So i know shop rando got added in 0.4.2 but looking around i think the only yaml option added was about prices, yes? Did you want to add that at all or just leave it at default? Otherwise LGTM

alwaysintreble:
the only new option is plando

PR #85, opened 2023-09-16
Add Terraria

Seldom-SE:
Idk what the game's weighting is supposed to be, so I just set it to the average, rounded to the nearest 10. You can change it or lmk what you want it to be


ScootyPuffJr1:
I've verified that this runs and the game weight seems fine.

PR #86, opened 2023-09-19
Update Pokemon Red and Blue.yaml

Alchav:
As yet untested

### DIFF
--- a/games/Pokemon Red and Blue.yaml
+++ b/games/Pokemon Red and Blue.yaml
@@ -1,127 +1,28 @@
 Pokemon Red and Blue:
-  accessibility: # Set rules for reachability of your items/locations.
-    #    Locations: ensure everything can be reached and acquired.
-    #    Items: ensure all logically relevant items can be acquired.
-    #    Minimal: ensure what is needed to reach your goal can be acquired.
-    locations: 0
-    items: 50
-    minimal: 0
-  game_version: # Select Red or Blue version.
-    red: 50
-    blue: 50
-    random: 0
-  trainer_name:
-    # Your trainer name. If not set to choose_in_game, must be a string not exceeding 7 characters.
-    # See the setup guide on archipelago.gg for a list of allowed characters.
-    choose_in_game: 50
-  rival_name:
-    # Your rival's name. If not set to choose_in_game, must be a string not exceeding 7 characters.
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
This is a lot of options. The game with the second most options has around 25 and this appears to have over 70. It would be great if you could trim some of them out. We only parse the options if they vary in at least two worlds. I'm not going to force you to get rid of any, but it would be nice.

Please get rid of the extraneous rows. Keeping this as clean as possible is preferred. I didn't try to run this, but I did notice death_link had a typo.

Alchav:
Trim out options?

ScootyPuffJr1:
I don't know this game at all, but I'm thinking things like starting_money. If you set that to a fixed value, we won't parse it out, which is how master_ball_price currently works. As long as every world gets the same value for an option, we don't have to care about that option. Do we need to allow for blind_trainers? Does confine_transform_to_ditto need to have a chance for false? The answers to these could easily be yes, but if some options could be set to always roll the same for all worlds, it would make things a little simpler.

Alchav:
I don't know what you mean by parse it out

ScootyPuffJr1:
After the async is generated, Berserker runs yaml_output_converter.py from this repo to create a csv file that I use to create the claiming spreadsheet. That csv file includes one record for each world with any options that make it different from other worlds. If an option is identical for all worlds for a given game, it won't show up in the csv. So if you set starting_money to 3000 for all worlds, that's one fewer option in the csv that I have to display on the spreadsheet. What I'm trying to say is that just because an option can vary in a yaml doesn't mean it has to for filler worlds.

Alchav:
Better?

ScootyPuffJr1:
Looks like this is down to about 50 or so options, which is still a lot but less than last async. It's up to Figment, but once the syntax errors are cleaned up, this is probably good to go.

t3hf1gm3nt:
for readability: please remove any lines with options that are set to their defaults. ones that i saw were elite 4 key item and pokedex conditions, confine transform to ditto, and i know there were others.
if an option only has one option but it is not the default option, that's fine to keep in

t3hf1gm3nt:
ran across a game appears unbeatable error with this. have posted a bug report to discord to investigate

PR #88, opened 2023-09-20
Added the Undertale yaml

jonloveslegos:


ScootyPuffJr1:
I've verified that this runs fine and the game weight seems good. We chatted in Discord about the option weights, so those seem fine, too.

PR #89, opened 2023-09-20
Update SC2 for 0.4.3

Ziktofel:
Updated to 0.4.3 YAML and set things for long async games accessible to new players

### DIFF
--- a/games/Starcraft 2 Wings of Liberty.yaml
+++ b/games/Starcraft 2 Wings of Liberty.yaml
@@ -1,19 +1,75 @@
 Starcraft 2 Wings of Liberty:
   game_difficulty:
-    casual: 20
-    normal: 30
+    casual: 25
+    normal: 35
     hard: 20
-    brutal: 10
-  upgrade_bonus: random
-  bunker_upgrade: random
+    brutal: 0
   all_in_map: random
   mission_order:
     vanilla: 10
     vanilla_shuffled: 30
     grid: 30
     mini_campaign: 30
+  final_map:
+    all_in: 25
+    random_hard: 75
   shuffle_protoss: random
   shuffle_no_build: random
   early_unit: true
   required_tactics: standard
   units_always_have_upgrades: true
+  generic_upgrade_research:
+    auto_in_no_build: 75
+    always_auto: 25
+  nco_items: true
+  bw_items: true
+  ext_items: true
+  mission_progress_locations: enabled
+  bonus_locations: enabled
+  challenge_locations:
+    enabled: 75
+    trash: 25
+  optional_boss_locations: enabled
+  triggers:
+    - option_category: Starcraft 2 Wings of Liberty
+      option_name: mission_order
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
I've verified this runs just fine, but I know nothing about the game to comment on option weights.

Ziktofel:
I'm the game's current maintainer so I know about these new stuff

The idea is to for full game enable all items and do reduction by sets only if a shorter campaign is in
For other settings I've picked the most popular options while still being friendly to new players (therefore no default Brutals)

games/Starcraft 2 Wings of Liberty.yaml
```diff
@@ -1,19 +1,75 @@
...
+  final_map:
+    all_in: 25
+    random_hard: 75
   shuffle_protoss: random
   shuffle_no_build: random
   early_unit: true
   required_tactics: standard
-  units_always_have_upgrades: true
\ No newline at end of file
+  units_always_have_upgrades: true
```
t3hf1gm3nt, line 20: just wondering if there should be a chance for this to be false? not saying it has to but would like some discussion about it

games/Starcraft 2 Wings of Liberty.yaml
```diff
@@ -1,19 +1,75 @@
...
+  generic_upgrade_research:
+    auto_in_no_build: 75
+    always_auto: 25
+  nco_items: true
+  bw_items: true
+  ext_items: true
```
t3hf1gm3nt, line 26: i did notice the triggers at the bottom that makes these have a chance to be false on all the mission orders besides pure vanilla. wouldn't it be smarter to put the chances here and have one trigger for vanilla to turn them all on? on that note, why the special exception for pure vanilla?

games/Starcraft 2 Wings of Liberty.yaml
```diff
@@ -1,19 +1,75 @@
...
+  mission_progress_locations: enabled
+  bonus_locations: enabled
+  challenge_locations:
+    enabled: 75
+    trash: 25
```
t3hf1gm3nt, line 31: i'd 50/50 this personally. interested in your reasoning behind the 3/4 split

games/Starcraft 2 Wings of Liberty.yaml
Ziktofel: Vanilla shuffled uses always on too, it's to not generate too many resources to the player

games/Starcraft 2 Wings of Liberty.yaml
Ziktofel: I think that most players want to have them enabled (the most usual setting) therefore this split

games/Starcraft 2 Wings of Liberty.yaml
Ziktofel: This setting to false is usually used on shorter games, therefore I kept that on

PR #90, opened 2023-09-20
Add DOOM 1993

ScipioWright:
Adding in Doom. I've discussed with Daivuk, the maintainer, already, and they gave general approval for this. The thing generates, and the options + triggers appear to work properly.

### DIFF
--- a/games/DOOM 1993.yaml
+++ b/games/DOOM 1993.yaml
@@ -1,0 +1,29 @@
+DOOM 1993:
+  difficulty: medium
+  random_monsters: random
+  random_pickups: shuffle
+  allow_death_logic: false
+  start_with_computer_area_maps: random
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
This seems good to merge.

games/DOOM 1993.yaml
```diff
@@ -0,0 +1,29 @@
+DOOM 1993:
+  difficulty: medium
```
t3hf1gm3nt, line 2: i would like to see chances for easy or maybe even baby. my gut instinct would break it down like
baby: 20
easy: 50
medium: 30
or if no baby, 50/50 between easy and medium

t3hf1gm3nt (approved):

PR #91, opened 2023-09-21
Created Stardew 4.x.x filler settings

agilbert1412:
@Dragorrod please do a sanity check to make sure I didn't do anything stupid

### DIFF
--- a/games/Stardew Valley.yaml
+++ b/games/Stardew Valley.yaml
@@ -65,14 +65,12 @@ Stardew Valley:
   multiple_day_sleep_enabled: 'true'
   multiple_day_sleep_cost: 0
   experience_multiplier:
-    vanilla: 20
     double: 50
-    triple: 20
+    triple: 40
     quadruple: 10
   friendship_multiplier:
-    vanilla: 20
     double: 50
-    triple: 20
+    triple: 40
     quadruple: 10
   debris_multiplier:
     vanilla: 25
@@ -80,7 +78,5 @@ Stardew Valley:
     quarter: 25
   quick_start: 'true'
   gifting: 'true'
-  mods:
-    []: 50
-    ["Bigger Backpack"]: 50
+  mods: []
   death_link: 'false'
--- a/games/Stardew Valley.yaml
+++ b/games/Stardew Valley.yaml
@@ -46,7 +46,11 @@ Stardew Valley:
     exclude_legendaries: 20
     exclude_hard_fish: 20
     only_easy_fish: 50
-  museumsanity: milestones
+  museumsanity:
+    none: 30
+    milestones: 50
+    randomized: 10
+    all: 10
   friendsanity:
     none: 25
     bachelors: 25
--- a/games/Stardew Valley.yaml
+++ b/games/Stardew Valley.yaml
@@ -1,34 +1,86 @@
 Stardew Valley:
   accessibility: items
-  starting_money: random-range-1000-3000
-  resource_pack_multiplier: 100
+  goal:
+    community_center: 60
+    grandpa_evaluation: 10
+    bottom_of_the_mines: 10
+    master_angler: 10
+    complete_collection: 10
+  starting_money: random-range-2000-20000
+  profit_margin:
+    normal: 30
+    double: 40
+    triple: 20
+    quadruple: 10
   bundle_randomization:
-    thematic: 2
-    vanilla: 1
+    vanilla: 10
+    thematic: 90
   bundle_price:
-    normal: 3
-    cheap: 1
+    very_cheap: 10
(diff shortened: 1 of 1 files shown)


games/Stardew Valley.yaml
```diff
@@ -1,34 +1,86 @@
...
+    true: 50
+  trap_items:
+    no_traps: 25
+    easy: 50
+    medium: 25
   multiple_day_sleep_enabled: 'true'
   multiple_day_sleep_cost: 0
-  experience_multiplier: 200
+  experience_multiplier:
+    vanilla: 20
```
Dragorrod, line 68: I think vanilla in both experience and friendship are bad ideas.   New players who did that in the past were full of regrets, and I would only keep the other options

games/Stardew Valley.yaml
```diff
@@ -1,34 +1,86 @@
...
-    half: 1
+    vanilla: 25
+    half: 50
+    quarter: 25
   quick_start: 'true'
-  gifting: 'false'
+  gifting: 'true'
+  mods:
+    []: 50
+    ["Bigger Backpack"]: 50
```
Dragorrod, line 85: I do not think having a mod is a good idea for the big async, since the goal is that new players can jump in.  The fact that they are not on the generic settings and mod setup is a seperate document makes the barrier to entry higher, and it should be avoided for newer players, in my opinion.

Dragorrod (approved):
Added 2 feedbacks, but I think it would be good as is too!

games/Stardew Valley.yaml
agilbert1412: Really? I thought this extremely minimal mod would be a great introduction to the concept, and in-game, the game tells you when connecting "you need backpack mod version X.x.x" so it's pretty explicit. But I can remove it if you want.

Dragorrod (approved):
Looks awesome

ScootyPuffJr1:
Yaml is clean and runs without issue. I can't comment on SDV content, but the syntax is good.

SnowOfAllTrades:
I still stand by my point that I brought up when Kaito proposed their original filler yaml: The entire gifting mechanic is against the spirit of the async and completely breaks progression chains. I strongly recommend turning it off for the fillers.

t3hf1gm3nt (commented):
wondering if there should be a chance of cropsanity, festival locations, and museum sanity being off. would keep the chance of it being on higher. maybe a 60/40 or even a 70/30 on/off split?
only other thing im wondering about for simplicities sake is setting one constant, higher amount of starting_money. like 5k or 10k. just spitballing, not entirely sold on changing it from what you have here.

agilbert1412:
Cropsanity: Hard disagree
Festivals: Soft disagree, willing to change it
Museumsanity: Agree, will make a change

t3hf1gm3nt:
care to explain your disagreements on cropsanity and festivals?
just want to understand your reasoning

agilbert1412:
Cropsanity in tandem with Seasons Randomizer are like 80-90% of the sources of BK in Stardew Valley. When one, the other, or both, get disabled, sphere 1 becomes massive, and the entire slot's progression curve becomes extremely "early-heavy" and leads to a much worse experience, as players then have way too much to do. This is especially true in such an async where people are expected to fully BK their seeds reasonably often, and can get hinted many things out of nowhere.

This causes some issues, like early items (some of which are requested in this very filler) being placed in worlds with way too many early locations, and in practice makes them not that early at all. This also causes more subtle issues, like a more difficult time trying to figure out what is and isn't in logic, an increased chance of missing early checks, and so on.

And on a more subjective note, these features are very core to the experience of this randomizer, as seeds and crops are core to Stardew Valley itself, so I feel like denying people who are possibly (probably?) first time players of these features makes for a bad marketing strategy and does not properly showcase the mod.


As for festivals, I am much less opiniated. The reasoning is simply that it is arguably the biggest new feature (4.x.x) that is newcomer friendly. Show up at the festivals, play the games, try the events (not even necessarily win!), participate in the community, get checks. Most of the other 4.x.x content is either geared towards veterans (perfection, ginger island, special orders) or for power users (mod support, but that's also why I wanted to add the backpack mod, as creating the yaml itself is the biggest obstacle to mod adoption. I thought it'd be nice to roll a few seeds with them, so even non tech-savvy people could try them). This feature (festivals) is very noob-friendly, and on easy difficulty, pretty casual, so I like it being on.

t3hf1gm3nt:
the festivals explanation is pretty reasonable
i admit im not 100% convinced about cropsanity, but this is mostly due to me not having much experience with it myself to be completely honest. but i'm willing to give it a shot with the current settings and see how people like it

PR #93, opened 2023-09-25
new Subnautica setting

SnowOfAllTrades:
adds the new setting "free samples" to the yaml, with a 50% chance of being true


t3hf1gm3nt:
@Berserker66 you good with these weights for the free samples setting?

Berserker66:
seems fine

PR #96, opened 2023-10-01
Fix a misspelled setting

SnowOfAllTrades:


ScootyPuffJr1:
They also dropped an s from randomize_weapon_level.

SnowOfAllTrades:
Of course they did. Fixed that too.

PR #102, opened 2023-10-09
Add Lufia 2 YAML

sidneymcdanger:
Weighted yaml for filler games for Lufia II. Focuses on longer games beatable by new players.

### DIFF
--- a/games/Lufia II Ancient Cave.yaml
+++ b/games/Lufia II Ancient Cave.yaml
@@ -41,11 +41,6 @@
     50: 4
     random-low: 2
 
-  death_link:
-    # When you die, everyone dies. Of course the reverse is true too.
-    false: 10
-    true: 1
-
   default_capsule:
     jelze: 1
     flash: 1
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
From MKI:

Iris treasure seems like it could be quite a bit too low, especially if it rolls less than 10/255, (YAY battling  3% chances on the fiirst iris treasure and then suffering through much lower odds for the rest of them, unless that's changed odds do get worst I'd recommend a minimum of 40 to air on a slightly easier side compared to the masochism that is 1/9 * 3/255 or 3/2295 that each normal chest turns into the desired last iris treasure, by the end of that world.

party start level should probably also be a range (1-10), it could be way too easy of random-low rolled level 50+ 


Blue chest:     random-high: 1 it has the issue of being too low on a chance (1/255), for reference I recall an earthbound LP that took a month for a 3/255 chance for the player too have RNG roll to get a Rare item.  Or the case where it can take years for a shiny Pokemon to appear for some people. 
Other than those two extremes I think it's fine.

sidneymcdanger:
Made updates based on feedback.

PR #105, opened 2023-11-18
Heavily reduce max science packs in factorio

Joethepic:
While having a filler seed for Factorio in the most recent async I don't have any techs under 500 and I think that is extremely unreasonable for newer players and for even older players considering mine went up to around 5500 packs for one check. While you can leave it in the background that is not as doable with biters and not everyone has a strong enough system to keep it running and play another game.

### DIFF
--- a/games/Factorio.yaml
+++ b/games/Factorio.yaml
@@ -21,7 +21,7 @@ Factorio:
   min_tech_cost:
     1: 50
   max_tech_cost:
-    random-range-low-100-10000: 20
+    random-range-100-1500: 20
   tech_cost_distribution:
     even: 10
     low: 10


t3hf1gm3nt:
@Berserker66 being the Factorio maintainer (and in general having better knowledge of the game than me): thoughts on this?

Berserker66:
The higher range of this option exists specifically for async(s)

eudaimonistic:
This is perhaps the single most reasonable and appropriate setting for the extreme option.  Players are not obligated to retain possession of these player slots (in our async anyway), and I'm pretty certain at least _someone_ would rise to the challenge as they did before.

ScootyPuffJr1:
Having the chance at a high cost is fine for the big async.

PR #106, opened 2023-11-20
Update A Link to the Past.yaml

Alchav:
- Update A Link to the Past yaml options to conform to the changes in this PR: https://github.com/ArchipelagoMW/Archipelago/pull/2357
- Move Pegasus Boots to `start_inventory_from_pool` (pls)

### DIFF
--- a/games/A Link to the Past.yaml
+++ b/games/A Link to the Past.yaml
@@ -15,13 +15,13 @@ A Link to the Past:
     on: 0 # prevents unshuffled compasses, maps and keys to be boss drops, they can still drop keysanity and other players' items
     off: 50
   ### End of Logic Section ###
-  bigkey_shuffle: # Big Key Placement
+  big_key_shuffle: # Big Key Placement
     original_dungeon: 50
     own_dungeons: 5
     own_world: 15
     any_world: 0
     different_world: 0
-  smallkey_shuffle: # Small Key Placement
+  small_key_shuffle: # Small Key Placement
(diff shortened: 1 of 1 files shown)


t3hf1gm3nt:
looks good so far but yeah we'll wait until that PR get's merged and in a release before we do anything with this

games/A Link to the Past.yaml
```diff
@@ -154,25 +145,45 @@ A Link to the Past:
...
+
+  randomize_shop_prices:
+    # Randomize the prices of the items in shop inventories
+    false: 50
+    true: 50
+
+  randomize_cost_types:
+    # Prices of the items in shop inventories may cost hearts, arrow, or bombs instead of rupees
+    false: 50
+    true: 50
```
Alchav, line 168:
```suggestion
    true: 5
```

eudaimonistic (approved):
Many of these changes are quite literally just porting options to their newer counterparts, and almost none of them are outright new/different values in the big picture sense.  Bombless start is a new one, and the 1/11 chance means we'll see at least some of it sprinkled into the async.  There isn't much risk in that option as players will inevitably get their weapons either way.

All changes look good to me, this is more 'modernizing the yaml' than it is making big changes to the gameplay output.  We also need this in order to even generate.

ScootyPuffJr1:
I'm good with everything except for maybe the keydrop shuffle. Are we sure we always want that on?

PR #107, opened 2024-01-08
Update Kingdom Hearts 2.yaml

shananas:
Removed options that no longer exists and added new options.
Only thing I'm unsure of is I chose to not hint weapon slot by default because its just makes the hints tab for KH2 kinda bad cause of how many hints you start with but since you don't start with these hints if you hint for something like flare force and it happens to be on a Donald staff that "costs 2 hints". In terms of Sora abilities all useful progression abilities cannot be on Keyblades so I do not see that as a problem. Only things would be Donald and Goofy limits that a player can be "punished because of this decision"

### DIFF
--- a/games/Kingdom Hearts 2.yaml
+++ b/games/Kingdom Hearts 2.yaml
@@ -15,11 +15,13 @@ Kingdom Hearts 2:
   Final_Form_EXP:
     4: 50
   Summon_EXP:
-    4: 50
+    6: 50
   Schmovement:
     level_1: 50
   RandomGrowth:
     0: 50
+  AntiForm:
+    'false': 50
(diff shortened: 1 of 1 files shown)


Silvris:
Fight difficulty could probably have a chance to roll easy, but I won't die on that hill.
Filler items local could go either way IMO. Other games like OoT don't local their filler items (and OoT has a chance of pot shuffle even). 
No real opinion on hint weapon slot.

eudaimonistic:
Weapon Slot hinting tends to throw a large number of hints into the pool.  Players do have the option to enable it in their custom slots, so it seems like good form/etiquette to not flood the pool with tons of hints right off the rip.

PR #108, opened 2024-03-25
Add KDL3 and Celeste 64

Silvris:
Weighting for the two not set in stone, but mainly focused on the weights given to similar games.


games/__meta__.yaml
```diff
@@ -1,19 +1,21 @@
...
 description: Archipelago Async Mystery Filler Settings
 requires:
-  version: 0.4.3
+  version: 0.4.5
 
 # Define your game here!
 game:
   A Link to the Past: 100
   Adventure: 7
+  Celeste 64: 15
```
eudaimonistic, line 10: Celeste 64 is an incredibly short game.  The game is free and pretty accessible, so even though 15 is a pretty high number for the game length and relative newness of it's release/inclusion, this seems reasonable enough as a starting point.

games/__meta__.yaml
```diff
@@ -1,19 +1,21 @@
...
   Dark Souls III: 31
   DLCQuest: 25
   Donkey Kong Country 3: 25
   DOOM 1993: 50
   Factorio: 31
   Hollow Knight: 88
   Kingdom Hearts 2: 55
+  Kirby's Dream Land 3: 25
```
eudaimonistic, line 18: I'm not seeing a strong representation in the currently submitted custom worlds (just above 440 as of writing) from this game.  Celeste 64 actually has double the submissions in comparison.  I would recommend lowering this one to 20 or lower.  There are a couple of other outliers in the meta settings in general, but I think a change here brings the game more in line with most of the other titles and submission trends.

eudaimonistic (commented):
I can't comment much on how playable the KDL3 settings are, but Celeste 64 seems good as is to me.  KDL3 might need slightly lower weights based on custom submissions trends.

PR #109, opened 2024-03-25
TUNIC yaml

ScipioWright:
Adding the TUNIC yaml. I'm getting feedback, but it's not the end of the world if this gets merged then I put up another PR to adjust it, eh?


ScipioWright:
Of note, I did not put any thought into the weight. I'm just going to assume that whoever is in charge will come up with a weight, so just putting a fairly default-feeling value.

PR #111, opened 2024-03-26
Update SC2

Ziktofel:
Tested with generating using params (SC2-only):
`--plando=items --multi=1000 --skip_prog_balancing --spoiler=0`


eudaimonistic (changes requested):
These triggers appear to no longer work in 0.4.5 due to not having their root options defined prior to triggers resolving.  When all of these are added back in, the yaml generates as intended.

Ziktofel:
Fixed

eudaimonistic (approved):

PR #112, opened 2024-03-26
Stardew Valley: Adapted filler yaml for 5.x.x

agilbert1412:
Added new options, changed the ones that have new values, try to make this make sense


ScootyPuffJr1:
I got some feedback from the Async Helpers. We think the CC goal might be weighted too high, especially with the possibility of minimum bundle price. Also, we think the weights for friendsanity none and bachelors might be better if swapped.

PR #117, opened 2024-03-26
Update A Link to the Past.yaml

eudaimonistic:
Apparently the prize shuffle option changed from single character options to full words, so this PR updates the `g` option to `general` accordingly.


eudaimonistic:
Turns out I'm dumb and this is already covered by #106

PR #118, opened 2024-03-26
Add new options for RoR2

kindasneaki:
Changes:
- Victory option
- Traps option
- Begin with loop no longer exists
- Lower the chance to get classic


eudaimonistic (approved):
I agree with reducing classic goal by 10.  Explore is generally the more defined randomizer experience at this point, and these weights at least make some of the original experience still available.  I don't think lowering the pickup step range by 1 is a significant enough change to be any kind of issue.

PR #119, opened 2024-03-27
Update SMW for 0.4.5

lilymnky-F:
Added New Options for SMW's release for AP 0.4.5


ScootyPuffJr1:
I think a couple of these options are a little harder than we are going for in the big async. I'd prefer if hidden moons and hidden 1ups were always off and I think the blocksanity weight should be lower.

lilymnky-F:
Removed Moon and Hidden 1up Checks, change blocksanity to 1/8

PR #120, opened 2024-04-01
Update Secret of Evermore.yaml

Silvris:
0.4.5 split sniffamizer into two options, and added a new result for it. The 50/50 on ingredients maps to the original 50/50 for on/full, while we give a 25% chance for the new option to be rolled.

### DIFF
--- a/games/Secret of Evermore.yaml
+++ b/games/Secret of Evermore.yaml
@@ -27,8 +27,11 @@ Secret of Evermore:
     on: 50
     full: 50
   sniffamizer: # Shuffles or randomizes drops in sniff locations
-    on: 50
-    full: 50
+    shuffle: 75
+    everywhere: 25
+  sniff_ingredients:  # Randomizes the item pool for sniff ingredients
+    vanilla_ingredients: 50
+    random_ingredients: 50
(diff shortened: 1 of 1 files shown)


games/Secret of Evermore.yaml
```diff
@@ -27,8 +27,11 @@ Secret of Evermore:
     on: 50
     full: 50
   sniffamizer: # Shuffles or randomizes drops in sniff locations
-    on: 50
-    full: 50
+    shuffle: 75
+    everywhere: 25
+  sniff_ingredients:  # Randomizes the item pool for sniff ingredients
+    vanilla_ingredients: 50
+    random_ingredients: 50
```
black-sliver, line 34: I don't think the new option should exist in fillers since it ~triples the total location count.

The following change would basically restore the old behaviour: 50% chance to use shuffled vanilla, 50% chance to use fully random ingredients
```suggestion
    shuffle: 100
  sniff_ingredients:  # Randomizes the item pool for sniff ingredients
    vanilla_ingredients: 50
    random_ingredients: 50
```

black-sliver (approved):
This should match what was in SoE before

PR #121, opened 2024-04-02
Add Pokemon Emerald

Zunawe:
I'm happy to give up control of this and just make recommendations or answer questions once it's in. There isn't a ton I have strong opinions about, and a lot of this is just guessing at what sorts of options are common and friendly.

One important note: Setting `badges` or `hms` to `shuffle` is the most expensive thing an Emerald yaml can do in a large async. I've seen one slot spend 10 minutes in `pre_fill` in a 1000-player multi. There's not much more I can do from the world code, but there are one or two PRs to core that can significantly speed it up. It's extremely unlikely to fail (I've never seen it fail). But spending an hour or more just on a handful of filler yaml `pre_fill`s is a lot. I don't think `shuffle` `badges` and `hms` would be missed too badly from filler yamls if it's decided the cost is too high.

I rolled this a few times to test and while working on other code, but haven't stress tested it yet.

### DIFF
--- a/games/Pokemon Emerald.yaml
+++ b/games/Pokemon Emerald.yaml
@@ -1,0 +1,297 @@
+Pokemon Emerald:
+  goal:
+    champion: 55
+    steven: 25
+    norman: 0
+    legendary_hunt: 20
+  badges:
+    vanilla: 0
+    shuffle: 30
+    completely_random: 70
+  hms:
+    vanilla: 0
+    shuffle: 30
+    completely_random: 70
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
Pokemon Emerald will need to be added to the meta file with a weight in order to get rolled.

eudaimonistic:
Emerald has a very large playerbase currently and would likely be justified in a meta weighting of 30 or higher, IMO.

PR #125, opened 2024-04-04
Add A Short Hike yaml

chandler05:
Add filler yaml for A Short Hike

### DIFF
--- a/games/A Short Hike.yaml
+++ b/games/A Short Hike.yaml
@@ -1,0 +1,12 @@
+A Short Hike:
+  goal: 
+    help_everyone: 50
+    fishmonger: 25
+    photo: 25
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
I have never played this game, but I have a concern with the possibility of golden_feather_progression rolling hard for these fillers. Should this be limited to easy and normal?

chandler05:
I don't think it's that big of an issue. The logic for A Short Hike with the golden_feather_progression is almost entirely a suggestion anyways, since you can bypass the need for any Golden Feathers regardless of the difficulty.

ScootyPuffJr1:
Works for me then. Can you add A Short Hike to the ___meta___ file with a weight you feel is appropriate?

PR #126, opened 2024-04-04
Update Super Mario 64 Settings for 0.4.5

RadzPrower:
NOTE: All weighted options from the previous version that were not evenly distributed have remained the same or are close to equivalent based on how stars are now percentage-based. Also, all weighted options other than `area_rando` utilize the 5-to-2 weights favoring easier and/or preferred settings that were present for weighted options in the prior version.

Notable changes include:
- Addition of Move Randomization as a weighted option. This is an all-or-nothing randomization rather than only some moves being randomized.
- Addition of a weighted option for keys to be progressive or not. This is largely a function of there being the new goal option of `all_bowser_stages` which means non-progressive keys can make basement key be a goal blocker under certain circumstances now. A more ideal solution might be a set of triggers for `all_bowser_stages` vs. `last_bowser_stage` but I'm only just really starting to get into triggers, so I'm not comfortable with setting that up just yet.
- Added randomized star counts for all potential star count options other than `stars_to_finish` (will address that separately). These are all randomized favoring the middle of their respective ranges which roughly correlates to their default values, but will provide a bit of variability.
- `stars_to_finish` has been changed to a `random-range-middle` such that the lower and upper ranges roughly correlate to the previous weighted option maximums and minimums but with the new percentage-based values. Being a middle range random value means it does not exactly match to the weights of the previous version, but being a `random-range` will provide a wider range of variety compared to the four separate values being ten apart from one another from the prior version.
- Assorted formatting changes such as moving single options to be in line with the option name and changing evenly weighted options to simply be random for conciseness.

### DIFF
--- a/games/Super Mario 64.yaml
+++ b/games/Super Mario 64.yaml
@@ -4,9 +4,7 @@ Super Mario 64:
     courses_only: 3
     courses_and_secrets_separate: 2
     courses_and_secrets: 1
-  progressive_keys:
-    true: 5
-    false: 2
+  progressive_keys: true
   enable_coin_stars:
     false: 5
     true: 2
@@ -38,7 +36,14 @@ Super Mario 64:
     items: 5
     locations: 2
   buddy_checks: random
-  exclamation_boxes: random
+  ExclamationBoxes: random
   completion_type:
     last_bowser_stage: 5
     all_bowser_stages: 2
+  triggers:
+    - option_category: Super Mario 64
+      option_name: completion_type
+      option_result: all_bowser_stages
+      options:
+        Super Mario 64:
+          progressive_keys: false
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
Everything here looks fine, but it would be very nice to have those triggers on the goal and progressive keys, as you mentioned.

Silvris:
```yaml
Super Mario 64:
  progressive_keys: true
  completion_type:
    last_bowser_stage: 5
    all_bowser_stages: 2

triggers:
  - option_category: Super Mario 64
    option_name: completion_type
    option_result: all_bowser_stages
    options:
      Super Mario 64:
        progressive_keys: false
```
Untested but should work.

RadzPrower:
Added Silvris' trigger and it appears to be working correctly when generating via the webhost (though it did take a few tries to get an "All Bowser Stages" goal to test). I'll hold this in draft for a little while longer in case anybody else has any contributions they want to make, but I am personally happy with how it sits currently.

ScootyPuffJr1:
I'm happy with how this looks with the triggers.

PR #127, opened 2024-04-06
Final Fantasy Mystic Quest and Bumper Stickers Filler YAMLs

megakidicarus:
Final Fantasy Mystic Quest and Bumper Stickers Filler YAMLs

### DIFF
--- a/games/Final Fantasy Mystic Quest.yaml
+++ b/games/Final Fantasy Mystic Quest.yaml
@@ -1,0 +1,110 @@
+Final Fantasy Mystic Quest:
+  progression_balancing: '0'
+  accessibility: items
+  logic:
+    friendly: 25
+    standard: 10
+  brown_boxes:
+    exclude: 10
+    include: 25
+    shuffle: 10
+  sky_coin_mode:
+    save_the_crystals: 25
+    shattered_sky_coin: 10
+  shattered_sky_coin_quantity:
+    mid_24: 10
+    high_32: 20
+  starting_weapon:
+    steel_sword: 50
+    axe: 50
+    cat_claw: 50
+    bomb: 50
+  progressive_gear:
+    'false': 20
+    'true': 25
+  leveling_curve:
+    normal: 5
+    one_and_half: 15
+    double: 30
+    double_and_half: 15
+    triple: 5
+    quadruple: 2
+  starting_companion:
+    none: 50
+    kaeli: 50
+    tristam: 50
+    phoebe: 50
+    reuben: 50
+  available_companions: '4'
+  companions_locations:
+    standard: 25
+    shuffled: 10
+  kaelis_mom_fight_minotaur: 'false'
+  companion_leveling_type:
+    quests: 30
+    quests_extended: 20
+    save_crystals_individual: 10
+    save_crystals_all: 5
+    benjamin_level: 10
+    benjamin_level_plus_5: 5
+  companion_spellbook_type:
+    standard: 25
+    extended: 10
+    random_balanced: 5
+  enemies_density:
+    all: 5
+    three_quarter: 10
+    half: 20
+    quarter: 10
+  enemies_scaling_lower:
+    quarter: 25
+    half: 5
+  enemies_scaling_upper: one_and_quarter
+  bosses_scaling_lower:
+    quarter: 25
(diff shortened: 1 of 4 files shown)


megakidicarus:
Many of the settings are somewhat loosely based/inspired by historical ALTTP weights, with some adjustments: FFMQ's "logic" is inspired off of item pool, "sky_coin_mode" is inspired by alttp's "goal[s]" featuring local MacGuffins (sky coin fragments in ffmq's case), and FFMQ's  "map_shuffle" is based off ALTTP's "entrance_shuffle" weights. 

non-specific "random [sic]" settings are based instead weighted for the purpose of spreadsheet design

Enemy weights are based on Final Fantasy 1's standard settings. 

I set "leveling_curve" weights to: potentially, be on the easier side, for a small amount slots.

games/Final Fantasy Mystic Quest.yaml
```diff
@@ -0,0 +1,110 @@
+Final Fantasy Mystic Quest:
+  progression_balancing: '0'
+  accessibility: items
+  logic:
+    friendly: 25
+    standard: 10
+  brown_boxes:
```
wildham0, line 7: in general "include" should be have a much higher weigh, since exclude/shuffle essentially remove these as ap locations, making for a lot less checks

games/Final Fantasy Mystic Quest.yaml
```diff
@@ -0,0 +1,110 @@
...
+    high_32: 20
+  starting_weapon:
+    steel_sword: 50
+    axe: 50
+    cat_claw: 50
+    bomb: 50
+  progressive_gear:
+    'false': 20
+    'true': 25
+  leveling_curve:
```
wildham0, line 25: normal and 1.5 xps are really slow for not really good reasons, i would suggest lower weight for these

games/Final Fantasy Mystic Quest.yaml
```diff
@@ -0,0 +1,110 @@
...
+    kaeli: 50
+    tristam: 50
+    phoebe: 50
+    reuben: 50
+  available_companions: '4'
+  companions_locations:
+    standard: 25
+    shuffled: 10
+  kaelis_mom_fight_minotaur: 'false'
+  companion_leveling_type:
```
wildham0, line 43: benjamin level+0 can be really punishing early game, i would switch it for benjamin+10

games/Final Fantasy Mystic Quest.yaml
```diff
@@ -0,0 +1,110 @@
...
+    regions_strict: 25
+    regions_keep_type: 25
+  doom_castle_mode:
+    standard: 25
+    boss_rush: 10
+    dark_king_only: 15
+  doom_castle_shortcut:
+    'false': 25
+    'true': 15
+  tweak_frustrating_dungeons:
```
wildham0, line 87: tweak dungeons is a standard flag, it's usually enabled by default, i would include false at a much lower weight

games/Final Fantasy Mystic Quest.yaml
```diff
@@ -0,0 +1,110 @@
...
+    'false': 25
+    'true': 15
+  map_shuffle:
```
wildham0, line 90: for map shuffle, overworld is the most accessible and should have the higher weighting, dungeons, overworld+dungeons and everything are advanced user settings and should much lower weight

wildham0 (commented):
i think that covers it from a beginner/intermediate player perspectives, there's some other unusual choices, but i'm not familiar enough with the big async format, so they probably make sense in that context

games/Final Fantasy Mystic Quest.yaml
megakidicarus: would outright removing be a recommendation?

games/Final Fantasy Mystic Quest.yaml
megakidicarus: I also increased 3x and 4x a bit.

Rooby-Roo:
As a Bumper Stickers player, there's basically no amount of settings you could change that would make the game "not beginner friendly". Pretty sure you could put every single setting on Random and have no issues. 
No traps is a good way to emphasize that, though.

PR #128, opened 2024-04-06
The Witness: Rewrite to modern options

Rooby-Roo:
Re-wrote The Witness YAML to accommodate the modern option names. No "functional" changes have been made, but the floor is open for ideas.

### DIFF
--- a/games/The Witness.yaml
+++ b/games/The Witness.yaml
@@ -1,5 +1,4 @@
 The Witness:
-  local_items: [Lasers]
   puzzle_randomization:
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
We have https://github.com/ArchipelagoMW/AsyncTools/pull/116 from Vi updating The Witness already. Should we close this and go with that one?

Rooby-Roo:
Didn't see that one. Whoops! Looks like that one has plans for what it wants to be doing as well, so that's cool.

PR Closed.

PR #130, opened 2024-04-07
Add Hylics 2 Filler YAML

mzbrubak:
I made a filler YAML for Hylics 2.  Please let me know if there's anything that I can change or do differently as this is my first time contributing to someone else's repository on GitHub.  I've included my reasoning for the chosen settings below:

Party Shuffle: 'true' is generally harder than 'false' so the setting is biased towards 'false'.  The chance for 'true' is still significant since this could also lead to an easier seed if you are sent party members earlier than you would get them in the base game.

gesture_shuffle: biased towards 'anywhere' as this is the main aspect of randomization in the game, and as is the case with party members, this can result in an easier seed by having access to powerful gestures early.  "tvs_only" is given a lower chance than "default_locations" as the difference between the two is not large and I recall anecdotal evidence that "tvs_only" could cause generation problems.

medallion_shuffle is biased heavily towards 'false' as some of the medallion locations are somewhat obscure

extra_items_in_logic is biased heavily towards 'true' since it ensures certain useful tools are required to finish the game.  The small chance of 'false' is included since these tools are by no means necessary to finish

random_start is forced to 'false' as the non-default starting locations result in a very small sphere 1

I should also note that I ran multiple test generations of this yaml alongside the yaml I'm submitting for the big async.  Out of 64 generations, only one encountered an error, and the error seemed to pertain to one of the other games.  I would be happy to conduct further test generations or other testing if required.


ScootyPuffJr1:
Can you please add Hylics 2 to the meta yaml so it has a chance to roll?

mzbrubak:
Thanks for reminding me to do that as well; Hylics 2 is now in the meta yaml

PR #131, opened 2024-04-07
Adding Doom II and Heretic

ScootyPuffJr1:


games/DOOM II.yaml
```diff
@@ -0,0 +1,20 @@
+﻿DOOM II:
+  difficulty:
+    baby: 2
+    easy: 5
+    medium: 3
+  random_monsters: shuffle
+  random_pickups: shuffle
+  allow_death_logic: false
+  start_with_computer_area_maps: random
```
Daivuk, line 9: if for first timers to AP, but know doom, I would put this `false`
if for frist doom timers, put `true`.
not random.

games/Heretic.yaml
```diff
@@ -0,0 +1,24 @@
+﻿Heretic:
+  goal: complete_all_levels
```
Daivuk, line 2: i would set that to boss levels only

games/DOOM II.yaml
ScootyPuffJr1: I matched what we have in Doom 1993 for maps, but I don't think it matters too much.

games/Heretic.yaml
Daivuk: ya it was how it was before, but the all boss levels are a kinda newish feature.
in the end, probably doest matter. Ship it.

PR #133, opened 2024-04-08
Add Zillion filler

mzbrubak:
I added a filler for Zillion.  Reasoning for the settings chosen described below.  I tested generation 22 times alongside my big async submission and the Hylics 2 filler I previously submitted and all generations were successful.  Please let me know if there's anything I should change or do differently as I'm still relatively new to Github (in particular, I'm not sure why the commits related to adding Hylics 2 are still shown when they've already been merged)

- The stats page suggests Zillion and Hylics 2 have similar play counts so I used the same weight in the meta yaml.
- The number of floppy disks required is kept at the default of 5, while the number of floppy disks available is randomly chosen between 5 and 8 and the number of red ID cards is randomly chosen between 1 and 3.  Since this means at least 6 items are required for go mode, it is unlikely that players will finish immediately, but if a player can clear all of their checks then they will have enough hint points to hint for their go mode.
- With the given skill distribution, 80% of seeds will be at skill 2 or lower; for reference, skill level 1 or 2 is at a comparable difficulty to the vanilla game.
- Gun_levels and jump_levels influence the stats of characters , effectively determining how many Zillions or Opa-Opas the players will need for max stats.  The 'restrictive' setting is not used since it prevents all characters from reaching max stats.
-  The room_gen option and randomize_alarms settings are both given a decent chance of being used; the skill setting accounts for the difficulty of jumps expected of the player so these settings should add a good amount of variety with not too much impact on difficulty.  The scope_early option is given a decent chance of being used as well, which mitigates any increased difficulty from randomize_alarms.


beauxq:
I'm not familiar with these async games, so the following should not be interpreted as me suggesting changes, but rather just giving information that might be relevant to someone that can be a better judge of what is desired.

- The `floppy_req` (number of floppies required) is one of the options that I personally like to vary the most, depending on how long of a game I'm looking for, and how much BK I want to deal with.
  - Note that the interaction between floppies available, and floppies required is safe. (If you say 6 floppies are required and 4 floppies are in the item pool, it will automatically increase the number in the item pool to 6)

- Most Zillion players prefer `room_gen` on, but it's off by default because it adds a bunch of extra time to generation (+ ~20 seconds for each player that has it on). So if someone doesn't care, it's better if it's off.
If you don't mind adding about 20 seconds to generation for each Zillion world, then the players might like it more if room gen is on more.

mzbrubak:
I updated floppy_req and floppy_disk_count to add more variance.  For now I set 3 as the least number of floppies needed to avoid jet seeds.  Beauxq raises a good point about room_gen; in my experience the time added is closer to 10 seconds, but that time cost can add up and I know generation speed is a limiting factor for these asyncs.  I don't suspect many Zillion fillers will show up in which case the cost of room_gen should be minimal, but I could see an argument for turning off room_gen to save time, and will do so if requested

PR #136, opened 2024-04-10
Add Yoshi's Island

PinkSwitch:
Not sure if this is good enough, tried to keep things interesting but still mostly random, but here's my comments;

shuffle-midrings- Kind of difficult setting, but in my observation a lot of people tend to enjoy it so I left a small chance to still roll it.

starting_world- High chance to get an early world, lower chance to get a late-game world, and very low chance to get a world with little progression (with level shuffle off)

Most other things were set to a random mid-long tier seed.

### DIFF
--- a/games/__meta__.yaml
+++ b/games/__meta__.yaml
@@ -57,6 +57,6 @@ game:
   Undertale: 45
   VVVVVV: 12
   Wargroove: 10
+  Yoshi's Island: 25
   Zillion: 7
   Zork Grand Inquisitor: 7
-  Yoshi's Island: 70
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
You need to add Yoshi's Island to the _meta_ file in order for it to roll any fillers.

PinkSwitch:
oh

games/__meta__.yaml
```diff
@@ -59,3 +59,4 @@ game:
   Wargroove: 10
   Zillion: 7
   Zork Grand Inquisitor: 7
+  Yoshi's Island: 70
```
ScootyPuffJr1, line 62: This is a very high weight, putting Yoshi's Island at the number 6 most popular filler game. About halfway through the custom submission process, only 6 people have submitted it, putting it at 33rd most popular. I think a weight of 25 is more appropriate. Also, you seemingly appended this game at the bottom of an alpha list.

PR #150, opened 2024-05-31
Remove chance of no bosses

ScootyPuffJr1:


Seatori:
I understand the intention of not wanting filler games that can be cleared by autocollect, but I don't think entirely removing the option to not require bosses is the only way of doing it.
My personal suggestion would be to add a trigger that, when 'none' is chosen as a boss requirement, it instead changes the world to require a number of Dragon Egg Shards that are forced to be local.
That way there's still gameplay involved in clearing the world, but bosses aren't strictly necessary for players who would prefer not to deal with them. Forcing them to be local also avoids the other issue of the world being cleared by autorelease instead.
Optionally, the Advancements which require bosses could also be excluded for this mode, to better emphasize the fact that players aren't needed to fight them. This is probably unnecessary though and could likely be achieved by just forcing the Dragon Egg Shard count to be higher than the number required.

t3hf1gm3nt:
Seconding Seatori's comment
set up a trigger for if required_bosses = none, set egg_shards_available/required to some random amount
and set the shards local like any other macguffin for fillers

PR #163, opened 2024-07-16
Added Mario & Luigi Superstar Saga

nmize1:
Meta yaml weight is arbitrary

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,6 +130,3 @@ dmypy.json
 
 # Outputted yaml file
 output/
-/.vs/AsyncTools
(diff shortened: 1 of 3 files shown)


ScootyPuffJr1:
This fails because of the normal option in randomize_bosses. I didn't look into any other potential failures, though.

nmize1:
Sorry, I'll get it fixed tonight, I generated a few with it but clearly not enough.

nmize1:
I must have tested it incorrectly before. Fixed several misnamed things, successfully generated 200 mlss games.

PR #165, opened 2024-07-16
output usable files for each game and rename mystery.yaml to weights.yaml

alwaysintreble:
this makes it so that GenerateMystery now outputs a valid file for each game in the output folder while it builds the weights file. also renames the output weights file to `weights.yaml` since that's what AP expects by default


ScootyPuffJr1:
Would it be possible to remove the last line of dashes, currently under the Zork options?

PR #167, opened 2024-07-19
Update Timespinner.yaml

ArashiKurobara:
Tweaked a lot of weights for the purposes of all-filler asyncs, though honestly there's really nothing _so_ wild in Timespinner that it shouldn't still work pretty well for fillers in ones with customs as well even if they're maybe a bit spicier than before. The weights I did use are still a bit arbitrary and I'm entirely open to feedback. (If nothing else I think the chance of unscaled boss rando needed to be ripped out, it looks like it was an oversight from the last yaml change with the option changing from rando off/on and scaling off/on separate to just off/scaled/unscaled as one option)

Also rearranged a lot of options to group similar options together (e.g. ones that add starting inventory, ones that add locations, ones that are largely logic-impacting), the TS yaml has kind of tended to just keep slapping new options at the end which makes some stuff kind of weird to parse IMO.

Testing was just a quick gen to make sure I didn't outright break anything since I'm not doing anything especially weird with this one.

### DIFF
--- a/games/Timespinner.yaml
+++ b/games/Timespinner.yaml
@@ -54,13 +54,7 @@ Timespinner:
   LootTierDistro: full_random
   ShowBestiary: true
   ShowDrops: true
-  TrapChance: 10
-  Traps:
-    - Meteor Sparrow Trap
-    - Poison Trap
-    - Chaos Trap
-    - Neurotoxin Trap
-    - Bee Trap
+  TrapChance: 0
     
   triggers:
     # swap Cantoran odds on an inverted start - it's an endgame boss location that can be as early as sphere 1 if inverted and quick seed are both true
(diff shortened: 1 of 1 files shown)


ArashiKurobara:
Realized that if I change the randoms to explicit 50/50s it's way easier to tweak them later even if it's no different immediately

ArashiKurobara:
Okay finally probably the last tweak, added a trigger to lessen the odds of an optional endgame sidequest boss check being there in conditions where it's highly likely to be in a the first couple spheres

ArashiKurobara:
Scipio's right, traps with instantaneous effects don't do much in an async, and a lot of the Timespinner ones can make getting through the tutorial real rough (which means if you come in late you might just be stuck in a loop of new game and dying to traps for a while...)

dafatman922:
Looks pretty good to me. I might go so far as to say it might be fine to let damage rando do something else (mostly buffs or mostly nerfs) just to keep it a bit spicy, but that's not necessary

ArashiKurobara:
Yeah, I considered small chances of mostly nerfs/buffs but I think to start I just wanna get damage rando in there at all

PR #169, opened 2024-07-19
Update Lingo settings

benny-dreamly:
I thought I'd take a look at some of the lingo settings and see what could be improved. used some feedback from the lingo channel as well. (ignore my dlc quest things, they didn't do anything)

### DIFF
--- a/games/Lingo.yaml
+++ b/games/Lingo.yaml
@@ -1,8 +1,8 @@
 Lingo:
   shuffle_doors:
-    none: 50
-    simple: 25
-    complex: 25
+    none: 33
+    simple: 33
+    complex: 33
   shuffle_colors: true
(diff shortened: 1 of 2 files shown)


benny-dreamly:
@hatkirby just so you know about the PR (I know Scipio pinged you in the channel in the discord)

games/Lingo.yaml
```diff
@@ -53,6 +55,8 @@ Lingo:
       options:
         Lingo:
           location_checks: reduced
+          progressive_orange_tower: false
```
hatkirby, line 58: There's no reason to add these two lines, because those options do nothing when door shuffle is off.

benny-dreamly: good point

PR #171, opened 2024-07-20
Update Hollow Knight.yaml

FlorianDeSmedt:
Adding some chances for harder settings, open for feedback.
Godhome ending would have a 1/25 chance with a 50/50 for regular godhome ending and godhome flower. Adding godtuner to local items since it's a mcguffin for godhome flower ending.
Path of pain: included would have a 1/5 chance.
Also adding some triggers for skips enabled, with a 2/3 chance for no skips, 1/6 for easy skips and 1/6 for medium skips.
The skips that are enabled with the easy category require a bit of knowledge but should be easily doable, the medium ones also require a bit of extra skill.

### DIFF
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -148,7 +148,7 @@ Hollow Knight:
         ProficientCombat: 'true'
         BackgroundObjectPogos: 'true'
         EnemyPogos: 'true'
-        ObscureSkips: 'true'
+        ObscureSkips: 'false'
         ShadeSkips: 'false'
         InfectionSkips: 'true'
         FireballSkips: 'false'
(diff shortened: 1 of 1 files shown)


dafatman922:
I think precise movement fits fine in easy (and above) instead of just medium, as well as proficient combat but I think that's more open to interpretation. I am also VERY unsure as to whether or not I think godhome ending should go in, but that is something that will have to be discussed

FlorianDeSmedt:
Lowered chance for POP, removed godhome again. Changed precise movement and proficient combat to 'true' in easy skips, moved shade skips to medium

SnowOfAllTrades:
I disagree with ObscureSkips in the Easy category, otherwise I completely agree with these changes.

FlorianDeSmedt:
Changed obscureskips to false for the easy category. Especially combined with precise movement it can give things that shouldn't be in easy skips

PR #172, opened 2024-07-21
Add game: A Hat in Time

Mysteryem:
AHit has a lot of options and many of them interact with other options. The interactions between most options have been controlled using triggers, which makes the yaml quite complicated. I have attempted to describe what is being done with comments.

The main important option is the trigger for controlling which DLCs are enabled and thus required for the player to own. I could not find any charts showing the DLC ownership percentages, so I took a guess at what might be common for the AP community.

Looking at Steam achievements, the most common DLC1 achievement is Prepare to Die at 12.6% of players, and the most common DLC2 achievement is Stick It To The Man at 16.3% of players. I anticipate that DLC ownership is higher in general (especially in the case of DLC1 because the achievement is more difficult) and higher yet again for the Archipelago community.

For filler-only asyncs, there is a trigger that allows harder options to remain enabled. Disabling this will replace all harder options with easier ones towards the end of the triggers.

Solo seed generation is set up under the expectation that https://github.com/ArchipelagoMW/Archipelago/pull/3663 gets merged, there are currently no safeguards in place in this yaml for the issues addressed in the PR that affect AP v0.5.0.

---

DLC requirements:
- No DLC: 60%
- DLC1 only: 7%
- DLC2 only: 8%
- Both DLC: 25%

Harder options for filler-only asyncs:
- LogicDifficulty is set such that `normal` is 10x as likely as `moderate`, which is 5x as likely as `hard`, and `expert` is disabled.
- There is a 5% chance of Death Wish being enabled when DLC1 is enabled, though usually with a limited number of contracts with `DWShuffle` being enabled at a ratio of 3:1. The maximum LogicDifficulty is reduced to `moderate` when Death Wish is enabled.

### DIFF
--- a/games/A Hat in Time.yaml
+++ b/games/A Hat in Time.yaml
@@ -1,0 +1,513 @@
+A Hat in Time:
+  accessibility: items
+
+  # Make all traps local when enabled.
+  # Non-local traps typically have little effect in an async, with most being triggered upon loading into the game and
+  # then being cleared upon dying in the spaceship hub area or entering a level.
+  local_items:
+    - Baby Trap
+    - Laser Trap
+    - Parade Trap
+
+  ### Trigger control to allow harder options for filler-only asyncs ###
+  trigger_harder_options:
+    'true': 100
+    'false': 0
+
+
+  ### DLC trigger control ###
+  # Players are probably going to have either no DLC or both DLC, though the DLC1 content is probably less popular to
+  # play compared to the DLC2 content.
+  trigger_dlc:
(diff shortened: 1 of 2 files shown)


Mysteryem:
Here is a spoiler log (no paths or playthrough) from a generation using 200 copies of mystery.yaml modified to always pick A Hat in Time:
[AP_84719271504320872445_Spoiler.txt](https://github.com/user-attachments/files/16326428/AP_84719271504320872445_Spoiler.txt)

PR #173, opened 2024-07-21
Tweak Filler Settings for LADX, SM64, and SMW

RadzPrower:
### Link's Awakening DX

- Made tradequest equally likely
- Explicitly limit instrument range to 6-8
- Changed all dungeon item randomization to be the same across each type. Weighting favors local placement but does introduce `any_world` and `different_world` chances as well.
- Made rooster evenly weighted
- Added entrance shuffle weights. 3 : 2 for dungeons and 3 : 1 for simple entrances.
- Turned on the added functionality to use boots if you hold the bracelet button
- Added a rare chance (1 : 10) for the music to be shuffled

### Super Mario 64

- Slightly increased the chance that entrances are shuffled
- Made move rando an even chance
- Made endless stair star range extend from 42-90% of total stars
- Gave an even chance of 1Up Boxes being included or not
- Increased the odds of `All Bowsers` goal so that it is now 3 : 5 in favor of `Last Bowser`

### Super Mario World

- Added Yoshi Egg Hunt as an equal opportunity goal and made all Yoshi Eggs local
- Added moons as a potential check at 1 : 2 odds
- Added secret 1ups and bonus blocks at 1 : 5 odds
- Increased the odds of blocksanity from 1 : 7 to 1 : 3
- Stun, literature, and thwimp traps were removed
- If blocksanity is rolled, Yoshi Egg count can then be rolled between 50 - 255 favoring the middle of that range. Also, stun, thwimp, and literature traps are reintroduced as local items.

### DIFF
--- a/games/Links Awakening DX.yaml
+++ b/games/Links Awakening DX.yaml
@@ -2,35 +2,47 @@ Links Awakening DX:
   logic:
     normal: 1
   tradequest:
-    false: 3
-    true: 2
+    false: 1
+    true: 1
   goal:
     instruments: 5
-    seashells: 2
+    seashells: 3
   instrument_count:
-    random-high: 1
+    random-range-6-8: 1
   trendy_game:
     easy: 2
     normal: 5
     hard: 2
   shuffle_nightmare_keys:
     original_dungeon: 5
-    own_world: 3
+    own_dungeons: 5
+    own_world: 5
+    any_world: 3
+    different_world: 1
   shuffle_small_keys:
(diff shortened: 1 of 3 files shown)


ScootyPuffJr1:
Super Mario 64 can fail to roll sometimes when the stars to finish is set high. I think we should bump the 90 down to 80.

RadzPrower:
I've ran these three specifically something like 50+ times with no failures, but maybe it's an issue with more worlds than more rolls.

PR #196, opened 2024-09-06
Changing Game Weights

ScootyPuffJr1:


DeamonHunter:
For anyone who is wondering: Total weight is 1695 (Unless I missed a value or 2).

IMO the numbers seem good, especially nerfs on the more problematic games from the last async.

PR #207, opened 2024-10-07
Make Floppies and Red ID Card Local

cjmang:
From an async perspective, this game is really gated by the Opa Opa Items and the Zillion Items.  Hunting for the Floppy Disks throughout an async multiworld isn't meaningful, because the only check gated by Floppies and the Red ID Card is the "Win" Check, whose sole purpose is to flag the world as cleared to AP.  To even get far enough to use Floppies and the Red ID Card requires enough Zillions and Opa Opas--so many in fact you can get every other check in the game.

Given that, making Floppies and the Red ID Card local means a player still has to acquire the right items to get all the necessary checks, they still have to hunt for a fair number of checks in their world due to needing the Floppy Disks, but they don't have to wait 2 weeks after otherwise clearing their world just to have the game formally cleared in AP.

### DIFF
--- a/games/Zillion.yaml
+++ b/games/Zillion.yaml
@@ -25,9 +25,9 @@
     4: 7
     5: 1
   starting_cards: random-high
-  room_gen:
-    false: 60
-    true: 40
+  map_gen:
+    none: 60
+    rooms: 40
   local_items:
     Floppy Disk: 1
     Red ID Card: 1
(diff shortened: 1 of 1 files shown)


ScootyPuffJr1:
This doesn't roll. It needs to be updated for AP version 0.5.1.

cjmang:
Can confirm it doesn't roll.  However it doesn't roll on master even without the changes in this PR.  I can take a look, but I'm fairly new to AP so I'm not sure what would've changed that would be breaking this.

cjmang:
@ScootyPuffJr1 I believe I've fixed the issue.

PR #208, opened 2024-10-21
Add local Dragon Egg Shards

Nidoskull:
...did nobody EVER add Dragon Egg Shards to this yaml?? Anyway, this *should* generate fine. Tested by generating a few worlds, didn't encounter any errors.

### DIFF
--- a/games/Minecraft.yaml
+++ b/games/Minecraft.yaml
@@ -11,11 +11,14 @@ Minecraft:
   include_postgame_advancements: off
   shuffle_structures: on
   structure_compasses: on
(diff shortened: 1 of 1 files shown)


dafatman922:
I'm going to be honest, idk why dragon egg shards have never been included.
That being said, there is a problem with this setup for the dragon egg shards: I think the odds of getting a world that requires ALL dragon egg shards is a bit too high (this happens anytime req > available). I think having a set number of available and just ranging the requirement is preferable

ScootyPuffJr1:
This was addressed in another PR.

PR #217, opened 2024-11-15
Add Yacht Dice yaml

spinerak:
Added Yacht Dice.

- Please check if the weight of 77 is OK, I chose it semi-randomly. Might be too high, but since Yacht Dice is free, fast and easy to learn, it might be fine. But you probably have a way to determine weights.
- Also I don't understand what the program outputs so please check if I did it correctly.....

### DIFF
--- a/games/Yacht Dice.yaml
+++ b/games/Yacht Dice.yaml
@@ -1,0 +1,72 @@
+﻿Yacht Dice:
+  game_difficulty:
+    easy: 20
+    medium: 50
+    hard: 50
+    extreme: 0
+  score_for_last_check:
+    random-high: 50
+  score_for_goal:
+    random-high: 50
+  minimal_number_of_dice_and_rolls:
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
I have concerns that the random-high used for score_for_goal will lead to worlds that are on the harder side. I haven't played Yacht Dice yet, so I'm purely basing this on the default of 777. Is this a valid concern? Should we use random-middle instead?

spinerak:
I don't think Yacht Dice will cause issues, it's a game that can be learned in 10 minutes. A high score_for_goal means you will need almost all of your items, while the default 777 requires maybe 70%-80% of your items.

What is the target audience for the settings? Are they familiar with Yacht Dice, or completely new? If they are familiar, then I see no problem at all. If they are completely new, the default yaml might be more fun, since the game is new too. The default yaml also already has many random elements (which items are in the pool).

PR #226, opened 2024-11-26
Update Timespinner.yaml

ArashiKurobara:
Updated option names to match 0.5.1 changes to future-proof (backwards-compatible for now but unsure if it's meant to stay backwards-compatible). Also removed explicit definitions for a couple minor options that were locked to their defaults anyway.

### DIFF
--- a/games/Timespinner.yaml
+++ b/games/Timespinner.yaml
@@ -1,6 +1,6 @@
 Timespinner:
   accessibility: full
-  local_items: [Timespinner Gear 1, Timespinner Gear 2, Timespinner Gear 3]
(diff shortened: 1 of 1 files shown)


ArashiKurobara:
Figment agrees it's not a bad idea to add the spindle to local, doesn't do anything itself (like the gears) and reduces the external requirement when a slot ends up logically looking for vanilla past access/back to the future to just the wheel instead of both pieces. (Wheel actually does do things, that one can stay "however it decides to shake out" =P)

Radagast81 (approved):

PR #241, opened 2025-01-16
Randomize shop slots for Hollow Knight

mail-liam:
This PR takes the existing shop allocations, leaves 1 guaranteed for each shop, and allocates the rest to `ExtraShopSlots`, which get distributed randomly to all shops.

### DIFF
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -95,16 +95,16 @@ Hollow Knight:
   RandomCharmCosts: -2
   MinimumEggPrice: 1
   MaximumEggPrice: 10
-  EggShopSlots: 5
-  SlyShopSlots: 8
-  SlyKeyShopSlots: 6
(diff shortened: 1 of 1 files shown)


dafatman922:
Async helper big's suggestion:

Allow the extra shop slots to be variable (suggestion: random-range-middle-30-50)

There is also the fact that Salubra technically has 2 shops, and as a result using this method will blost Salubra a little bit. This can be resolved by not using extra shop slots, and just randomizing each individual shop's count. I'll leave it up to you whether or not that's something you want to do

mail-liam:
I'm not sure how well the item/location balance is for this HK - is it an issue if the low end is rolled in that case?

I hear you on Salubra, but I'm fine to run it with these settings once to get some data on if it's an _issue_ per se. Salubra already has the most checks in vanilla, so it's not that far from the baseline experience.

PR #242, opened 2025-01-16
Enable costsanity for Hollow Knight

mail-liam:
Vanilla cost weights have been left as-is. They can be tweaked later with player feedback.
Considered adding a 3/1/1 split of off/shops_only/on,, but figured I'd leave it simple for now.

### DIFF
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -109,14 +109,15 @@ Hollow Knight:
   SplitMothwingCloak: 'random'
   SplitMantisClaw: 'random'
   CostSanity:
-    off: 3
+    off: 4
+    shops_only: 1
     on: 1
   CostSanityHybridChance: 10
   CostSanityEggWeight: 1
-  CostSanityGrubWeight: 1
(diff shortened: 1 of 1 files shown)


dafatman922:
Async helper big here

This looks alright, everything I will say is mostly up to you;

Suggestion: decrease the weight of eggs specifically, because they are a bit unintuitive, and maybe also slightly decrease the weight of geo.

As well, I like the idea of adding a bit of shops_only. Your specific ratio brought up also tackles my other suggestion which is increasing the possibility of costsanity. 4/1/1 might be a bit better though

mail-liam:
Sounds good to me! I wavered between 1/3 and 1/4 originally. so that seems like a good compromise (I know it's preferable to aim a little higher to increase multiple option rolls on a given slot).

I doubled the chances of every other currency, which puts the egg chance at ~4.3%. LMK if you think that's ok to start with.

mail-liam:
Actually maybe I should just adjust the whole cost options to be out of a pool of 100 to make it easier to reason about.

PR #244, opened 2025-01-16
KH2 - Add Aerial Recovery to start inventory

Nidoskull:
This was discussed in the Dec/Jan Post Async Discussion thread.


Nidoskull:
whoops this is a dupe of #238 actually

PR #247, opened 2025-01-18
Update Pokemon Red and Blue.yaml

ArashiKurobara:
As of 0.5.1 trainersanity is no longer all or nothing; tweaked the trigger for "both dex- and trainersanity" to make the trainersanity most likely partial. Weights and range bounds modeled off the dexsanity values (full trainersanity is 317 so did similar proportions out of 317 as dexsanity does out of 151).

Additionally removed blind trainers values from the triggers as trainers with checks now ignore the setting entirely, making the option there superfluous.

### DIFF
--- a/games/Pokemon Red and Blue.yaml
+++ b/games/Pokemon Red and Blue.yaml
@@ -177,7 +177,6 @@ Pokemon Red and Blue:
         Pokemon Red and Blue:
           trainersanity: full
           dexsanity: disabled
-          blind_trainers: 0
           exp_modifier: 
             16: 50
             32: 25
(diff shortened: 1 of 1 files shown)


ArashiKurobara:
... actually I realized right after making the pull that dexsanity used to be a percentage and is now a count, so the trigger as it was was probably 30%-100% range (i.e. ~45 to 151 Pokemon) rather than 30-100 Pokemon.

On the other hand, since the range options only appear in the trigger that turns *both* sanities on, I think keeping the range dialed back a little bit kinda works? They do both still have the chance to roll full even in that trigger anyway.

PR #257, opened 2025-01-24
TUNIC: 0.6.0 options

ScipioWright:
Additional Combat Logic adds logic to most areas for stat upgrades, magic weapons, and other stuff like that. I figure pushing it to this rate is probably fine for now since if someone doesn't want it and there's no slots without it, they wouldn't mind taking one with it since they can just play as if it was off anyway.

I put the Grass Rando line there because I'm pretty sure you don't want it, but if you do want it, it's easier to adjust it. With grass rando on and local_fill set to default, it'll place 95% of a tunic grass rando world's filler items across tunic grass rando worlds. That 5% left is ~300 items, which will be mostly grass. And local_fill can be increased up to 98% to further reduce grass in the pool to ~120 items per world. But also it spams the fuck out of the text client and I could see that being annoying.
So, up to you there.

### DIFF
--- a/games/TUNIC.yaml
+++ b/games/TUNIC.yaml
@@ -1,7 +1,7 @@
 TUNIC:
   # column order for the spreadsheet:
   # Start with Sword | Keys Behind Bosses | Hexagon Quest | Gold Hexagons Required | Percentage of Extra Hexagons | Entrance Rando | Fewer Shops | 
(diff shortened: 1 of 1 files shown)


dafatman922:
big approves

dafatman922:
I don't know how much you're worried about the sheet order, but breakable shuffle is not included in your list at the top. Otherwise this is fine

PR #260, opened 2025-02-23
Removal of MC and StS

ScootyPuffJr1:


ScootyPuffJr1:
This will not happen unless these worlds are removed from core.

PR #261, opened 2025-03-03
Aquaria - Change skip_first_vision

palex00:
There is no reason for it to be false. It just saves time.

### DIFF
--- a/games/Aquaria.yaml
+++ b/games/Aquaria.yaml
@@ -50,7 +50,7 @@
 
   skip_first_vision: true
 
-  local_items: ["Energy form"]
+  local_items: ["Energy Form"]
   
   triggers:
     - option_category: null
--- a/games/Aquaria.yaml
+++ b/games/Aquaria.yaml
@@ -48,7 +48,7 @@
     'false': 1
     'true': 1
 
-  skip_first_vision: false
+  skip_first_vision: true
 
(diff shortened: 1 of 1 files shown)


tioui:
When I created the file, ScootyPuffJr1 told me that the runs did not need to be fast. So, I did not try to save time. But then again, I don't have anything against it either.

tioui (approved):

ScootyPuffJr1:
Yaml does not roll. Seems to be a problem with Energy Form.

tioui:
> Yaml does not roll. Seems to be a problem with Energy Form.

That's... quite strange. This option should not change anything related to Energy Form. I will try to check what's not right (but tomorow; tonight, I can't).

games/Aquaria.yaml
```diff
@@ -48,7 +48,7 @@
     'false': 1
     'true': 1
 
-  skip_first_vision: false
+  skip_first_vision: true
 
   local_items: ["Energy form"]
```
tioui, line 53: Since the next Big Async will use Archipelago 0.6.0, it will be necessary to change "Energy form" by "Energy Form" (capital F)
```suggestion
  local_items: ["Energy Form"]
```

palex00:
@ScootyPuffJr1 It is fixed, may I request re-review?

PR #265, opened 2025-03-15
[HK] Remove THK and Any goals

Nidoskull:
Discussed around here: https://discord.com/channels/731205301247803413/1346222534034591867/1346238369096663131

THK goal is, skill-wise, identical to the Sealed Siblings goal; requiring the player to defeat The Hollow Knight within the Black Egg. However, THK requires three less items to goal (namely the forced-local King_Fragment, Queen_Fragment and Void_Heart). `any` goal is functionally an `all` goal, which means achieving the THK ending is still a viable goal. Additionally, it also activates Grub Hunt goal, allowing for potentially unexpected goalings.

Removing THK would not do much without also removing `any`, as `any` still allows for THK goals.

### DIFF
--- a/games/Hollow Knight.yaml
+++ b/games/Hollow Knight.yaml
@@ -67,13 +67,11 @@ Hollow Knight:
   RemoveSpellUpgrades: 'false'
   StartLocation: king's_pass
   Goal:
-    any: 2
-    hollowknight: 3
     siblings: 3
     radiance: 3
+    grub_hunt: 2
     godhome: 0
     godhome_flower: 0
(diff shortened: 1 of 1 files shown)


dafatman922:
With the removal of THK and Any, the weight of Siblings should go up to balance out the ratio of "not Radiance" goals to Radiance goals. I would say 8:3:2 for Sealed Siblings, Radiance, and Grub Hunt respectively, is good (it's the same as it was before, if you count Any as "not Radiance"). I would vouch a higher chance of Radiance if the next big async wasn't going to include customs

PR #273, opened 2025-04-03
Slight changes to game weights

ScootyPuffJr1:
I lowered the weight of the games that didn't really get claimed in the last async and took a guess with a couple new games.


Nidoskull:
Unsure on Saving Princess's weight being 8, but it can always be adjusted later. LGTM otherwise.

PR #277, opened 2025-04-07
Add egg option to Witness yaml

dafatman922:
Currently it has a 50/50 of being enabled. If enabled, a somewhat arbitrary distribution of difficulties is given

### DIFF
--- a/games/Minecraft.yaml
+++ b/games/Minecraft.yaml
@@ -1,7 +1,7 @@
 Minecraft:
   goal:
     advancements: 1
-    eggs: 1
+    eggs: 2
   combat_difficulty:
     easy: 10
     normal: 20
@@ -25,7 +25,7 @@ Minecraft:
         Minecraft:
             advancement_goal: random-range-60-80
             required_bosses:
-                none: 30
+                none: 20
                 ender_dragon: 20
                 wither: 20
                 both: 10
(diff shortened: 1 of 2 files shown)


dafatman922:
Ignore all the other commits this PR dragged in with it, I don't understand how github works. There is only one file change associated with this PR and it's the one I describe

NewSoupVi:
Egg Hunt is a seasonal feature that is visible & default on in April, and invisible & default off outside of it

The omission of it in the yaml is entirely intentional, so that it would be on for all slots in a big async generated in April and off for all slots outside of it

Whether that's good or desired is not ultimately my call to make, but it was not a mistake

NewSoupVi:
One thing I could see is to have two different sets of weights, something like

In April weights:
20% off, 20% easy, 40% normal, 20% hard

Outside of April weights:
70% off, 10% easy, 10% normal, 10% hard

And then you have a trigger where if `easter_egg_hunt` is `off`, it rerolls using the "outside of april" weights, and if it's `normal`, it rerolls using the "inside of April" weights

NewSoupVi:
Is this "is_it_april" something that's already convention, like in a meta.yaml or something? Like someone is gonna have to make sure to change that right?

ScootyPuffJr1:
I advised big to force one async as if it were April, just in case this generation actually takes place in May. I will make the necessary changes after this async is done.

NewSoupVi (approved):
In that case, I approve :) Not that my approval was necessary but

PR #295, opened 2025-07-17
[meta] More Yacht Dice

g0goTAS:
The people demand more Yacht Dice, so we gotta deliver.

Seriously though, every Yacht Dice was picked in 8 hours. A weight of 30 is way too low, so let's set it to 100 instead.


ScootyPuffJr1:
I'll handle game weighting in a little while, but I won't go this extreme.

PR #296, opened 2025-07-17
Update mission count & selection for the sa2b yaml

Newsupermarcobro:
I noticed that the mission count and the mission select was still using the old settings so I updated them to the newer settings and just kept the weights equivalent to what they would be.


ScootyPuffJr1:
Near duplicate of https://github.com/ArchipelagoMW/AsyncTools/pull/298. Which is better?

ScootyPuffJr1:
Closing as duplicate

PR #297, opened 2025-07-20
[Civ VI] Excluded Boost Space Race and Urbanization

Gearwcu:
Boosts don't have an era set they should be achievable by.  So logic assumes these can be completed in the ancient era which they can't be.
There may be more but these locations have a chance to be needed to send via console, especially the space port which is maybe possible from modern era after 5 progressive eras.
Issue:
https://github.com/ArchipelagoMW/Archipelago/issues/5136

### DIFF
--- a/games/Civilization VI.yaml
+++ b/games/Civilization VI.yaml
@@ -21,3 +21,11 @@ Civilization VI:
       options:
         null:
           name: CivVI-{player}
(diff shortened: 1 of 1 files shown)


Gearwcu:
added 2 More boosts affecting Eras + boosts
issue: https://github.com/ArchipelagoMW/Archipelago/issues/5210

PR #310, opened 2025-08-30
[AsyncTools] + [The Witness] Create custom override for month sensitive options

rocket0634:
This change is an evolution of the previous "is it april" option that is now automated based on the month AsyncTools is run, as opposed to changing it manually.

The month can still be changed manually if desired in the option result of the trigger. It can be changed to match the current month for the April option, or changed to not match for the default option.


ScootyPuffJr1:
Magic

PR #313, opened 2025-08-31
[Subnautica] Update yaml to 0.6.3

dafatman922:
Give this at least a day to cook because I am not fully confident on the decisions made in this yaml.

The concept is that it'll be more difficult to day one cheese the Infected goal with the empty_tanks setting on, and because one of the players is really adamant about wanting to play Infected, I think it is reasonable to try it. If early goals remain an issue for Subnautica, we can turn it off again.

What I am less certain on is what the setting should do for the OTHER goals. I've decided at the moment that free will also always have empty tanks on (it's also a very cheesable goal), and make it random for specifically Launch goal. I could see a very good argument for just having empty tanks on always, but I don't know if the community wants that


dafatman922:
Made the decision to just have empty_tanks always on. If someone has an issue with empty_tanks they can feel free to voice it and I will probably change it for the next one, but it's not actively harmful in any way

PR #319, opened 2025-09-04
SC2 0.6.4 update

Ziktofel:
Updates the SC2 YAML for 0.6.4

### DIFF
--- a/games/Starcraft 2.yaml
+++ b/games/Starcraft 2.yaml
@@ -12,7 +12,10 @@ Starcraft 2:
     wol: 10
     hots: 10
     lotv: 10
-    big_async_canvas: 5
+# Commented out as bugged by following bugs:
+# https://github.com/ArchipelagoMW/Archipelago/pull/5604
+# https://github.com/ArchipelagoMW/Archipelago/issues/5605
+#    big_async_canvas: 5
   mission_order:
     # See also trigger handling race swaps
     vanilla: 5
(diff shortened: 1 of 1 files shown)


games/Starcraft 2.yaml
```diff
@@ -1,27 +1,35 @@
...
+    mini_campaign: 10
+    gauntlet: 10
+    golden_path: 15
+    hopscotch: 15
   game_difficulty:
     casual: 25
     normal: 35
     hard: 20
     brutal: 0
+  difficulty_damage_modifier: random
```
Snarkie, line 29: This means that every filler generated has a 50% chance to have

Ziktofel: It's only for Brutal/Casual and in the campaigns it's used in vanilla.

True is vanilla difficulty-based behavior (depending on campaign/mission)
False is the modifier is always shut off (done in the mod by overriding the user data)

The default is true (vanilla behavior), there's no support for enabling that modifier in for example in WoL (the map triggers for example disable it for the duration of the cutscenes)

games/Starcraft 2.yaml
```diff
@@ -182,38 +432,44 @@ Starcraft 2:
...
+          victory_cache: random-range-0-3
     - option_category: Starcraft 2
       option_name: maximum_campaign_size
       option_result: 40
       options:
         Starcraft 2:
           excluded_missions:
             - Supreme
-            - The Infinite Cycle
+            - The Infinite
```
MatthewMarinets: "The Infinite" looks like a typo here.

Ziktofel:
Blocked by: https://github.com/ArchipelagoMW/Archipelago/pull/5604

Ziktofel:
The test generation was successful after the patch which is the PR blocked by. (tested 32 parallel gens, 100 worlds each, all successful)

ScootyPuffJr1:
I'm ready to start generating. What needs to happen to get this to work the way you want (or at least good enough for now)?

ScootyPuffJr1:
I'm not going to wait for a bugfix to roll this next async. What specific parts of this yaml don't work? Can they be removed/commented out? If this isn't in a mergeable state soon, I will consider rolling without SC2.

PR #388, opened 2026-04-02
Allow Zork and SADX to gen

rocket0634:
This probably works on Scooty's machine this is mostly just to fix an oversight on the repo.


Evereliquest (approved):
Had to change this myself too.

PR #389, opened 2026-04-02
Add Mario Kart Double Dash

rocket0634:
I'm a bit disappointed in the lack of Shorter Courses and having all 16 courses be included in the all cup tour, but given that the Big Async is a month long event, having a bit of a longer goal isn't that much of a problem.

The full three lap all cup tour tends to take over half an hour and will need to be reattempted for each failed attempt (though it is very easy to not fail if you're in go mode)

I haven't tried 200cc myself so I can't comment on it.

Other than those things I don't have any issues with the YAML. I'll just go ahead and submit what Bunk posted in the Double Dash channel, and I decided to give DD a weight similar to Outer Wilds.

### DIFF
--- a/games/Mario Kart Double Dash.yaml
+++ b/games/Mario Kart Double Dash.yaml
@@ -1,0 +1,102 @@
+﻿
+Mario Kart Double Dash:
+
+  goal:
+    all_cup_tour: 50
+    trophies: 30
+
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
Included in https://github.com/ArchipelagoMW/AsyncTools/pull/400

PR #394, opened 2026-04-04
Add EarthBound stuff

PinkSwitch:
i am tired and didn't want to set a dev env up so i kinda ooga booga'd this together manually
yaml and triggers were tested and confirmed to be working
please yell at me if I forgot something or something is incorrect


ScootyPuffJr1:
This fails when shop_randomizer hits shuffled.

PR #403, opened 2026-05-03
Update satisfactory trigger to use intermediate value

adamsong:
When Archipelago runs `get_choice` it will check if there is a randomization, and roll it. When one of the goal triggers was setting `goal_selection` it was setting it to an array, which if there was another goal triggers, would be randomly pulled from. Adding the dummy option of `goal_selection_for_trigger` allowed setting a weighted value that the triggers can act on to write to `goal_selection` and allow setting it to an array without that array being randomized.

I tested this, and all 4 possible goal options show up as the correct goal in the spoiler log, and successfully generate.

### DIFF
--- a/games/Satisfactory.yaml
+++ b/games/Satisfactory.yaml
@@ -20,11 +20,12 @@ Satisfactory:
     10: 30
     25: 25
     50: 15
-  goal_selection:
+  goal_selection_for_trigger:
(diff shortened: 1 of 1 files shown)


Jarno458:
I approve this, i did try the yaml triggers, and they seem to work fine now

Jarno458 (approved):

PR #404, opened 2026-05-15
Linux Misc Fixes

FlitPix:
This PR contains small fixes for using GenerateMystery.py on Linux.

- Filenames, including extensions, are case-sensitive on Linux. Bumper Stickers's yaml now ends with `.yaml` (lowercase), since using `.YAML` (uppercase) caused [`open(f"games/{safe_name}.yaml", encoding="utf-8-sig")`](https://github.com/ArchipelagoMW/AsyncTools/blob/main/GenerateMystery.py#L83) to fail.
- The terminal color is now reset on exit.


duckboycool (approved):
Haven't tested, but changes look correct to me. I think it's a little weird how the control codes are being concatenated to things that are already f-strings, but I guess that was already present anyway.

FlitPix:
i wouldn't mind changing it to a new line if that's what you prefer. don't really care either way

PR #405, opened 2026-05-29
Remove local Letter from Pokemon Emerald yaml

andyoam:
Local letter removed avenues for Emerald players to escape Dewford Jail as there are other means of doing so that were still distributed through the multiworld.  Letter unwalls were byfar the rarest means of escaping Dewford due to this.

### DIFF
--- a/games/Pokemon Emerald.yaml
+++ b/games/Pokemon Emerald.yaml
@@ -199,7 +199,7 @@ Pokemon Emerald:
     true: 20
   death_link: false
   enable_wonder_trading: true
-  local_items: Letter
+  local_items: []
 
   triggers:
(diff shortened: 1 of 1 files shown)


games/Pokemon Emerald.yaml
lammyson, line 181: Can we consider increasing the `true` weight so there's more of a chance out of Dewford Jail?

Nidoskull: Out of scope for _this PR specifically_, but yeah, we can do that.

lammyson: Just saw this PR (#411) which looks good!

lammyson (approved):

PR #406, opened 2026-05-29
Celeste (Open World) - Adjust filler amount

dafatman922:
Due to complaints about too many raspberries, the possible number of strawberries required has been increased dramatically. Since the goal is percentage-based, this should have relatively little impact on the amount of "stuff" the player has to do / get.

Additionally, to make the percentage requirement more accurate to what is posted, the junk_fill_percentage is now 0


games/Celeste (Open World).yaml
```diff
@@ -82,6 +83,12 @@ Celeste (Open World):
           roomsanity:
             "false": 1
             "true": 3
+    - option_category: Celeste (Open World)
+      option_name: roomsanity
+      option_result: "true"
+      options:
+        Celeste (Open World):
+          total_strawberries: random-range-400-500
```
etnpce: The cap for total_strawberries is 202, so this will always be capped.
I think this should be ~120-202

dafatman922: 120-202 doesn't solve the issue at all. I will have to rethink this, thank you for informing me

dafatman922:
I'm not fully content with where we landed with these settings, roomsanity just causes too many problems.

That said, if we don't find a better idea, this is fine to merge. Let's not let perfect be the enemy of good

PR #409, opened 2026-05-30
[Emerald] Un-local the Letter

Nidoskull:
After feedback, this was Not A Good Move. It *might* have gone over better if people actually read changelogs, but without an Early Letter option, this feels more restrictive to people.

(Note for pedants: This does not *force* the Letter to be non-local, just makes it not forced local again like it was.)


Nidoskull:
whoops i didn't see #405

PR #412, opened 2026-05-30
[Pokemon Crystal] Make Ruins of Alph tiles local for unown_hunt goal

lammyson:
Make Ruins of Alph tiles local for unown_hunt goals

### DIFF
--- a/games/Pokemon Crystal.yaml
+++ b/games/Pokemon Crystal.yaml
@@ -1653,6 +1653,7 @@ Pokemon Crystal:
     mild: 9
     wild: 1 # palnote: mercy.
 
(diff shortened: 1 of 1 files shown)


palex00 (changes requested):
As the person who made the original YAML, this seems like a good change. However, you don't need to do this trigger-stuff. The tiles are only added in Unown Hunt, so simply putting them as local should just work.

lammyson:
> As the person who made the original YAML, this seems like a good change. However, you don't need to do this trigger-stuff. The tiles are only added in Unown Hunt, so simply putting them as local should just work.

Done!

PR #419, opened 2026-06-03
[SC2] update yaml

Snarkie:
Updates to the SC2 yaml, based on observed issues and player feedback.

## Add new slot with no NCO ##
Apparently, many players own the original campaigns, but have never bought the Nova Covert Ops mission packs. We had multiple requests for a slot, that doesn't include NCO but is not limited only to the free to play campaigns.

## Increase weights for bigger slots ##
Traditionally, bigger slots are in higher demand for the big async. We don't want to over-do it, but the weights for bigger slots get a slight increase.

## Bigger small/medium slots ##
The smaller slots did feel a little short, and especially the challenge/any_unit slots had trouble getting enough checks for a good enough composition. Increasing the mission count for smaller slots by roughly 20% across the board.

## Reduced locations for bigger slots ##
This was planned from the start, but never made it into the yaml. By reducing the checks per mission for bigger slots, we hope to reduce excessive filler items and better balance the big slots against other games.

## Weapon & Armor upgrades
Challenge slots on any_units logic do not handle W/A upgrades very well. They are very important to keep up with the enemies in harder missions, so this is a logic issue we will address in a future update. Fix this for now by forcing enough W/A upgrades into the pool.

## Reduce Hopscotch, increase Blitz ##
For mission orders, Hopscotch has received negative feedback for being too linear and restrictive, while Blitz was received more positively, despite being at a much lower weight. Reducing Hopscotch and increasing Blitz weights.

## Redundant Keys ##
This is a planned option for a future SC2 version, but some slots can already take advantage of it. 
- Big Async logo slots now add 2 additional redundant copies of each key into the pool
- Swap of the Race slots now add 1 additional redundant copy of each key

Additionally, there is some formatting changes, and the file was saved with an option to cull trailing whitespaces, which causes some of the differences.

----

This yaml was tested by generating several 50 mission multiworlds. No generation failure of any kind was observed.

### DIFF
--- a/games/Starcraft 2.yaml
+++ b/games/Starcraft 2.yaml
@@ -3,12 +3,14 @@ Starcraft 2:
     # Overarching trigger, selecting one of these formats
     # Each format should consistently roll the exact amount of missions specified
     # This should provide a way to easily identify them in the options table
-    standard: 25
-    # 50 missions, Grid/Golden Path/Hopscotch/Blitz
-    challenge: 15
-    # 40 missions, any_unit logic, reduced checks, always hard/brutal, trap items
-    free_to_play: 20
-    # 35 missions, only f2p (WoL, Prophecy, Prologue)
+    standard: 18
+    # 60 missions, Grid/Golden Path/Hopscotch/Blitz
+    challenge: 10
+    # 50 missions, any_unit logic, reduced checks, always hard/brutal, trap items
+    free_to_play: 13
+    # 45 missions, only f2p (WoL, Prophecy, Prologue)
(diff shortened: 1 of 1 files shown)


MatthewMarinets (approved):
My gut says the swing from hopscotch/standard to blitz is quite large (blitz doubles in chance, which seems a bit much when the description implied the feedback was that hopscotch in particular was too constrained). But that's a small sub-branch of the trigger tree, so we can see how it goes.

Everything else looks good. Did a quick local test to confirm that locking keys works, and it did for me.

PR #420, opened 2026-06-03
[SC2] change output splitter to display custom SC2 option

Snarkie:
Just a quick&dirty modification that allows displaying a custom option in the formatted options list using the big async options splitter.

grafik

`Slot Type` is a trigger in our yaml, which determines many of the aspects of the slot and serves as a good summary or orientation. It is intended to convey the information of multiple other settings in one row, something that many other games could probably benefit from as well.
(well... the trigger is actually called `game format`, but I feel like Slot Type better captures the meaning for player-facing info)

This is not ready to merge, I am sure, this can be improved upon. Should probably be more generalized, so other games can interface with it.

### DIFF
--- a/async_sheet_tools/csv_output_splitter.py
+++ b/async_sheet_tools/csv_output_splitter.py
@@ -34,6 +34,7 @@ def format_output_csv(filename:str = "output.csv", output_filename:str = "format
                 # get all the actually rolled options and their index in the list
                 # the index is used to filter for the relevant options later
                 all_options_indices = {option_name:index for index, option_name in enumerate(row)}
+                all_options_indices[len(row)] = "Slot Type"
             else:
                 game_name = row[1]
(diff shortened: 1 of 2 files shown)


ScootyPuffJr1:
I disagree with this fundamentally. The csv splitter should be game agnostic.

Snarkie:
> I disagree with this fundamentally. The csv splitter should be game agnostic.

exactly, this is why this PR is in draft. Basically just me trying if it works at all. I do agree that a final version should work with other games, I'm just not sure, how that should work/how to implement that exactly.

I do think the "Slot Type" option could be a general column available to all games, and if they include it in their whitelist, it would show up. Then we just need some sort of generalized interface games can use to provide whatever information each slot should display in this column. And a way to map from options that are actually present in the table to the slot type.

PR #423, opened 2026-06-04
[Waffles] Avoid the issues we experienced last time

NyanMC:
The issue with Golden Yoshi Eggs was supposedly (and hopefully) fixed. Is this overcompensating? Maybe, but it should still mean more people get to play more of their game before goaling even in a worst case scenario.

Changes:
- Made Bowser goal more likely than Yoshi's House
- Removed `swap_level_exits` as it never did anything and doesn't exist anymore. Intended behavior was incorporated into a 50/50 chance for `swap_exit_count` to roll 0
- Added more traps with temporary effects to local items
- Reduced upper bound on Golden Yoshi Egg counts for blocksanity

### DIFF
--- a/async_sheet_tools/gameoption_whitelist.yaml
+++ b/async_sheet_tools/gameoption_whitelist.yaml
@@ -1643,7 +1643,6 @@ Slay the Spire:
     - Map Teleport Shuffle
     - Map Transition Shuffle
     - Carryless Exit Count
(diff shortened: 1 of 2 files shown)


NyanMC:
updated the PR to remove Swap Level Exits from the option whitelist because it literally does not exist anymore and that may otherwise cause issues

PR #430, opened 2026-06-06
[Nine Sols] Shop Rando/Root Node rando and logic difficulty weight adjustment

AliciaWhimsicott:
These two options have been out for a while now and have shown no issues with logic for multiple months. This sets both of those options to 50/50s. It also changes the vanilla:medium logic ratio from 3:1 to 2:1 as tricks in that logic difficulty are not very challenging to learn.


AliciaWhimsicott:
This PR now also updates the whitelist file to add the two new options to the whitelist, as well as the Logic Difficulty option.

PR #432, opened 2026-06-06
[Dark Souls 3] Adding items to the starting inventory

JDario16:
To avoid soft logic walls I added these items in the starting inventory.

Doing this people don't have to wait for them and it could avoid 80% of the game being in the same sphere


ScootyPuffJr1:
I don't want to add progression items to starting inventory.

PR #433, opened 2026-06-07
GsTla yaml tweaks

Dragion147:
- Disabled NPC scaling to improve generation times
- (Community Feedback) Enable Manual Retreat Glitch flag, QoL to use a glitch that can be used otherwise but is tedious
- (Community Feedback) Increase Odds to start with the ship to give a more open world feeling


Snarkie (approved):
Simple changes as discussed in the community thread. Approving.

PR #441, opened 2026-06-13
[Stardew] Ajust goals, bundles per rooms, and backpack size

ploki122:
Adjusted goal weight
Allowed 1 and 2 backpack size
Allowed rolling more extra bundles per room.


ploki122:
@agilbert1412 Any concern over these changes?

agilbert1412:
I think the extra bundle amounts are a bit much, but not like insane. Its probably fine

agilbert1412 (approved):

PR #444, opened 2026-07-03
[Pokemon Black and White] Add early_parcel triggers

lammyson:
[Pokemon Black and White] Add early_parcel triggers

### DIFF
--- a/games/Pokemon Black and White.yaml
+++ b/games/Pokemon Black and White.yaml
@@ -95,7 +95,7 @@
   early_parcel:
     yes_please: 50  # Add Parcel to start_inventory
     no_thanks: 40   # Parcel can be anywhere
-    maybe_worse: 10 # Add Parcel to local_items. Can be behind Surf/Strength/Basement Key
+    maybe_worse: 10 # Add Parcel to local_items. Can be behind Surf/Strength/Basement Key/Cut
 
   triggers:
     - option_category: Pokemon Black and White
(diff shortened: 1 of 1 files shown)


andyoam:
Check quantities were posted prior via gergeshwan that indicate that B/W is able to hint with even no sanity options following receiving the Dragon Skull item.  In this case, adding Parcel as a local item increases the size of Sphere 1 from 24 to 52 checks, while Loot Sack adds an additional 12 checks and Dragon Skull adds an additional 27.

Due to this, I don't think there's any reason to entertain the ongoing discussion in the thread about including Machine Part at all in this equation.

As far as the starting items/local items/starting hint proposals, I don't think these need to be handled as triggers since they're pretty small changes from seed to seed.  I do not, however, think that any of these should be handled as local items.  This causes a coinflip scenario where either the BW player does not interact with the multiworld basically at all for the first 1/6-1/4 of their game or they get walled behind 5 potential unwalls that they're unable to hint out of.  This creates an even more hostile situation a la the local Letter for Emerald.

That leaves Starting Items and Starting Hints.

I would propose Parcel and Loot Sack as Starting Items and Dragon Skull as a Starting Hint.  If we wanted to include triggers, we could have combinations of the three meshed between Starting Items and Starting Hints, but like I said above, I don't think the difference between them is large enough to necessarily justify that.  I know Scooty doesn't like starting hints but some worlds implement them baseline and there's clearly issues with how B/W behaves with the big async so short of removing it, implementing a starting hint seems the best avenue.

PR #448, opened 2026-08-09
DC1 rebalance

Lizardman175:
- Removed all_bosses settings; it's a bit silly in the BA
- Rebalanced settings based on early slot claims


Lizardman175:
As an added bonus, turning off `all_bosses` should have a big affect on gen time.  Some quick tests ran in about 78% of the time off rather than on.

PR #449, opened 2026-08-09
[Terraria] Remove Calamity

NyanMC:
Due to recent Calamity drama (which I will not explain here), I get the feeling that not as many people are going to want to play Calamity in the next big async. ~~This pull request reduces its weight from 50% to 20%. I have opted to not turn off Calamity entirely in the event that there are still people that want to play it. This weight change may be overkill, or perhaps it's not enough; it should probably be discussed in the post-async thread before merging this.~~

Update 2026-08-10:
This pull request now prevents Calamity from rolling completely.

### DIFF
--- a/async_sheet_tools/gameoption_whitelist.yaml
+++ b/async_sheet_tools/gameoption_whitelist.yaml
@@ -2126,7 +2126,6 @@ Terraria:
     - Terraria
     - Locations
     - Name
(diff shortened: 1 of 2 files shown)


NyanMC:
i didn't realize this hadn't been merged yet and deleted the branch early. my bad

